PR #111, opened 2021-10-19
More explicit batchWriter errors

Aschen:
## Description

Since the `document:create` action does not necessarily take an ID, when one of the document cannot be created it was more difficult to identify which one was errored in the `errors` array.

A dirty fix solution was to use the `strict: true` mode for the `mCreate` action. The problem is that if I create 3 documents and only one cannot be created, the 3 promises will be rejected (even if 2 document had been created) and this make debug even more difficult.

This PR allows to reject only the promise of the document who was effectively in error by comparing the document content returned by Kuzzle

### Example

For the following code:
```
import { BatchWriter, BatchController } from './lib/core-classes/BatchProcessing';
import { Kuzzle, WebSocket } from 'kuzzle';

const kuzzle = new Kuzzle(new WebSocket('localhost'));

kuzzle.on('reconnected', () => console.log('reconnected'))
kuzzle.on('networkError', () => console.log('networkError'))

async function run () {
  await kuzzle.connect()

  const writer = new BatchWriter(kuzzle);
  writer.begin();
  const doc = new BatchController(kuzzle, writer)
  try {
    await Promise.all([
      doc.create('test', 'test', { name: 'Aschen'}),
      doc.create('test', 'test', { age: 42}),
      doc.create('test', 'test', { age: 'Aschen'}),
    ])
  }
  catch (error) {
    console.error(error)
  }
}

run();
```

Error before:
```
MultipleErrorsError: At least one of the mCreate actions failed
```

Error after:
```
BadRequestError: Cannot create document in "test":"test" : failed to parse field [age] of type [integer] in document with id '8GiqnnwBnIIUvXsZ-P47'. Preview of field's value: 'Aschen'
```

### DIFF
--- a/lib/core-classes/BatchProcessing/BatchController.ts
+++ b/lib/core-classes/BatchProcessing/BatchController.ts
@@ -67,7 +67,7 @@ export class BatchController extends DocumentController {
     const error = errors.find(({ _id: id }) => id === _id);
 
     if (error) {
-      throw new BadRequestError(`Cannot replace document "${index}":"${collection}":"${_id}": ${error}`);
+      throw new BadRequestError(`Cannot replace document "${index}":"${collection}":"${_id}": ${error.reason}`);
     }
 
     return successes[idx];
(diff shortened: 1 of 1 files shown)


Aschen:
@rolljee Done :+1:

PR #113, opened 2021-10-20
removed device-create from documentation as it is not supported anymore

### LINKED ISSUE #75: [doc] Delete device:create
It is no longer supported to manually create a device via the API endpoint (which is no longer available).

rolljee:
Remove `device:create` from the documentation

close #75


UncleSamSwiss:
For some reason this commit isn't reflected in the published documentation yet:
https://docs.kuzzle.io/official-plugins/device-manager/1/controllers/device/create/

This is confusing as the latest published version does no longer have this method.

What is the right way to create a device? Is provisioning the only way?

rolljee:
@UncleSamSwiss 

This is because the documentation isn't reflecting the dev branch but the master and the documentation hasn't been updated yet.
And yes provisioning is the way to go

PR #117, opened 2021-10-25
Allow to skip asset historization

Aschen:
## Description

Add another event before historization, this event has a `historize` boolean property (`true` by default) and if this property is set to false then the asset will not be historized


Aschen:
I close this PR until we finished the work on the event system.

PR #129, opened 2021-10-28
move logic of registerDecoder inside the decodersController & service

### LINKED ISSUE #37: Move Decoders from init of the plugin to the dedicated DecodersService
# Description

Decoders are initialized at the root level of the plugin (inside the init fonction) it will be better to move all that logic of init at the place where it belong -> DecodersService

- [x] DecoderService / DecoderController (#114)

- [x] Add an accessor to the PluginContext in the Decoder base class

rolljee:
close #37

### DIFF
--- a/lib/DeviceManagerPlugin.ts
+++ b/lib/DeviceManagerPlugin.ts
@@ -210,9 +210,7 @@ export class DeviceManagerPlugin extends Plugin {
     this.api['device-manager/asset'] = this.assetController.definition;
     this.api['device-manager/device'] = this.deviceController.definition;
     this.api['device-manager/decoders'] = this.decodersController.definition;
-    
-    console.log(this.api);
-    
+
     this.pipes = {
       'device-manager/device:beforeUpdate': this.pipeCheckEngine.bind(this),
       'device-manager/device:beforeSearch': this.pipeCheckEngine.bind(this),
--- a/lib/DeviceManagerPlugin.ts
+++ b/lib/DeviceManagerPlugin.ts
@@ -308,7 +308,7 @@ export class DeviceManagerPlugin extends Plugin {
   private getPayloadsMappings (): JSONObject {
     const mappings = JSON.parse(JSON.stringify(this.config.adminCollections.payloads));
 
-    for (const decoder of this.decoders.values()) {
+    for (const decoder of this.decodersService.decoders.values()) {
       mappings.properties.payload.properties = {
         ...mappings.properties.payload.properties,
         ...decoder.payloadsMappings
--- a/lib/DeviceManagerPlugin.ts
+++ b/lib/DeviceManagerPlugin.ts
@@ -204,10 +204,15 @@ export class DeviceManagerPlugin extends Plugin {
     this.decodersController = new DecodersController(this, this.decodersService);
     this.engineController = new EngineController('device-manager', this, this.deviceManagerEngine);
 
+    const payloadControllers = this.decodersService.getPayloadController(this.payloadService);
+
+    this.api['device-manager/payload'] = payloadControllers;
     this.api['device-manager/asset'] = this.assetController.definition;
     this.api['device-manager/device'] = this.deviceController.definition;
     this.api['device-manager/decoders'] = this.decodersController.definition;
-
+    
+    console.log(this.api);
+    
     this.pipes = {
       'device-manager/device:beforeUpdate': this.pipeCheckEngine.bind(this),
       'device-manager/device:beforeSearch': this.pipeCheckEngine.bind(this),
@@ -219,8 +224,6 @@ export class DeviceManagerPlugin extends Plugin {
     await this.initDatabase();
 
     await this.migrationService.run();
-
-    this.decodersService.registerPayloadController(this.api, this.payloadService);
   }
 
   /**
--- a/lib/DeviceManagerPlugin.ts
+++ b/lib/DeviceManagerPlugin.ts
@@ -204,9 +204,7 @@ export class DeviceManagerPlugin extends Plugin {
     this.decodersController = new DecodersController(this, this.decodersService);
     this.engineController = new EngineController('device-manager', this, this.deviceManagerEngine);
 
-    const payloadControllers = this.decodersService.getPayloadController(this.payloadService);
-
-    this.api['device-manager/payload'] = payloadControllers;
+    this.api['device-manager/payload'] = this.decodersService.buildPayloadController(this.payloadService);
     this.api['device-manager/asset'] = this.assetController.definition;
     this.api['device-manager/device'] = this.deviceController.definition;
     this.api['device-manager/decoders'] = this.decodersController.definition;
--- a/lib/DeviceManagerPlugin.ts
+++ b/lib/DeviceManagerPlugin.ts
@@ -102,11 +102,7 @@ export class DeviceManagerPlugin extends Plugin {
     return this.context.accessors.sdk;
   }
 
-  /**
-   * List of registered decoders.
-   * Map<model, decoder>
-   */
-  private decoders = new Map<string, Decoder>();
+  private decoders = [];
 
   private deviceMappings = new DeviceMappingsManager(devicesMappings);
 
@@ -198,8 +194,8 @@ export class DeviceManagerPlugin extends Plugin {
     this.batchWriter.begin();
 
     this.payloadService = new PayloadService(this, this.batchWriter);
-    this.deviceService = new DeviceService(this, this.decoders);
     this.decodersService = new DecodersService(this, this.decoders);
+    this.deviceService = new DeviceService(this, this.decodersService.decoders);
     this.migrationService = new MigrationService('device-manager', this);
     this.deviceManagerEngine = new DeviceManagerEngine(this, this.assetMappings, this.deviceMappings);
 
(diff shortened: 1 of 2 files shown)


lib/DeviceManagerPlugin.ts
```diff
@@ -247,24 +198,12 @@ export class DeviceManagerPlugin extends Plugin {
...
-  ): { controller: string, action: string } {
-    decoder.action = decoder.action || Inflector.kebabCase(decoder.deviceModel);
-
-    if (this.api['device-manager/payload'].actions[decoder.action]) {
-      throw new PluginImplementationError(`A decoder for "${decoder.deviceModel}" has already been registered.`);
+  ) {
+    if (! this.decodersService) {
+      this.decoderQueue.push({ decoder, handler });
+    } else {
+      return this.decodersService.register(decoder, { handler });
```
Aschen, line 205: This shouldn't be possible and should raise an exception. Even if the decoder is register, the controller action will not be added to Kuzzle

lib/DeviceManagerPlugin.ts
```diff
@@ -381,4 +320,53 @@ export class DeviceManagerPlugin extends Plugin {
 
     return request;
   }
+
+  private getDefaultConfig () {
```
Aschen, line 324: If you really want to extract this then you could put it in a separate file for example

lib/core-classes/DecodersService.ts
```diff
@@ -1,31 +1,70 @@
...
+  ): { controller: string; action: string } {
+    decoder.action = decoder.action || Inflector.kebabCase(decoder.deviceModel);
+
+    if (this.api["device-manager/payload"].actions[decoder.action]) {
+      throw new PluginImplementationError(
+        `A decoder for "${decoder.deviceModel}" has already been registered.`
+      );
+    }
+
+    this.api["device-manager/payload"].actions[decoder.action] = {
```
Aschen, line 55: This break the single responsibility principle, this class should be aware of other classes internal mechanisms (here it's the plugin API definition from the main Plugin class).

We can discuss about possible solutions if you want

lib/core-classes/DecodersService.ts
```diff
@@ -1,31 +1,70 @@
...
-  constructor(plugin: Plugin, decoders: Map<string, Decoder>) {
+  constructor(plugin: Plugin, decoders: Map<string, Decoder>, decodersQueue: DecoderQueue[], payloadService: PayloadService) {
     this.config = plugin.config as any;
     this.context = plugin.context;
+    this.api = plugin.api;
+    this.payloadService = payloadService;
 
     this.decoders = decoders;
+
+    decodersQueue.forEach(({ decoder, handler }) => {
```
Aschen, line 29: Try to avoid `forEach` loops because they uses function which gave no control on the execution flow.

For example this will not wait for each decoder but rather run all the functions in //
```
      decoders.forEach(async decoder => {
        await someFunction();
        this.register(decoder);        
      })
```

lib/core-classes/DecodersService.ts
```diff
@@ -1,31 +1,70 @@
...
+      this.register(decoder, { handler });
+      this.context.log.info(`Register API action "device-manager/payload:${decoder.action}" with decoder "${decoder.constructor.name}" for device "${decoder.deviceModel}"`);
```
Aschen, line 31: This log line should be done when the decoder is registered on the plugin by the user otherwise it's more difficult to debug the stacktrace when a faulty decoder has been registered

lib/core-classes/DecodersService.ts
```diff
@@ -1,31 +1,70 @@
...
+    });
   }
 
   async list(): Promise<DecoderContent[]> {
-    const decoders = Array
-      .from(this.decoders.values())
-      .map(decoder => decoder.serialize())
+    const decoders = Array.from(this.decoders.values()).map((decoder) =>
```
Aschen, line 36: Nitpicking
```suggestion
    const decoders = Array.from(this.decoders.values()).map(decoder =>
```

lib/core-classes/DecodersService.ts
```diff
@@ -1,31 +1,70 @@
...
+    decoder.action = decoder.action || Inflector.kebabCase(decoder.deviceModel);
+
+    if (this.api["device-manager/payload"].actions[decoder.action]) {
+      throw new PluginImplementationError(
+        `A decoder for "${decoder.deviceModel}" has already been registered.`
+      );
+    }
+
+    this.api["device-manager/payload"].actions[decoder.action] = {
+      handler: (request) =>
```
Aschen, line 56: Nitpicking
```suggestion
      handler: request =>
```

lib/core-classes/DecodersService.ts
```diff
@@ -1,31 +1,70 @@
...
+        handler
+          ? handler(request, decoder)
+          : this.payloadService.process(request, decoder),
+      http: decoder.http,
+    };
+
+    this.decoders.set(decoder.deviceModel, decoder);
+
+    return {
+      controller: "device-manager/payload",
```
Aschen, line 66:
```suggestion
      controller: 'device-manager/payloadt,
```

package.json
```diff
@@ -8,9 +8,9 @@
     "url": "https://github.com/kuzzleio/kuzzle-plugin-device-manager"
   },
   "scripts": {
+    "docker:up": "docker-compose up",
```
Aschen, line 11: This script has the same name in every other repository. To preserve consistency it should stay like this or be changed in every other repository (I'm not even sure it's a good idea since our users are used to this scrip)

You can rather create an alias if you want

lib/DeviceManagerPlugin.ts
```diff
@@ -82,6 +81,7 @@ export type DeviceManagerConfig = {
 
 export class DeviceManagerPlugin extends Plugin {
   public config: DeviceManagerConfig;
+  private decoderQueue: DecoderQueue[] = [];
```
Aschen, line 84: This property name and type name are misleading since a [Queue](https://www.wikiwand.com/en/FIFO_(computing_and_electronics) is a type of data structure but here you just use this as a simple array

Aschen (changes requested):

lib/core-classes/DecodersService.ts
```diff
@@ -1,31 +1,64 @@
...
   async list(): Promise<DecoderContent[]> {
     const decoders = Array
-      .from(this.decoders.values())
+      .from(this._decoders.values())
       .map(decoder => decoder.serialize())
 
     return decoders;
   }
+
+  printDecoders() {
```
Aschen, line 46: IMHO could be private

lib/core-classes/DecodersService.ts
```diff
@@ -1,31 +1,64 @@
...
+    for (const decoder of this._decoders.values()) {
+      this.context.log.info(`Register API action "device-manager/payload:${decoder.action}" with decoder "${decoder.constructor.name}" for device "${decoder.deviceModel}"`);
+    }
+  }
+
+  register(decoder: Decoder) {
```
Aschen, line 52: Missing ts doc

lib/core-classes/DecodersService.ts
```diff
@@ -1,31 +1,64 @@
...
+    this.decoders.set(decoder.deviceModel, decoder);
+  }
+
+  registerPayloadController(api: PluginApiDefinition, payloadService: PayloadService) {
```
Aschen, line 56: It's better in term of architecture but still I'm not completely satisfied with this solution for 2 reasons:
  - this class have to modify another class internal state (`api` object)
  - the controller definition can not be used somewhere else (for example in an custom PayloadController with other actions)

lib/core-classes/DecodersService.ts
```diff
@@ -1,31 +1,64 @@
-import { PluginContext, EmbeddedSDK, Plugin } from "kuzzle";
+import { PayloadService } from './PayloadService';
```
Aschen, line 1: Usually we sort the import like this:

1. nodejs import ( `fs`)
2. external lib (`kuzzle`)
3. internal lib (`./lib/...`)

Aschen (changes requested):

lib/DeviceManagerPlugin.ts
```diff
@@ -204,10 +204,15 @@ export class DeviceManagerPlugin extends Plugin {
...
 
+    const payloadControllers = this.decodersService.getPayloadController(this.payloadService);
+
+    this.api['device-manager/payload'] = payloadControllers;
     this.api['device-manager/asset'] = this.assetController.definition;
     this.api['device-manager/device'] = this.deviceController.definition;
     this.api['device-manager/decoders'] = this.decodersController.definition;
-
+    
+    console.log(this.api);
```
Aschen: Debug leftover

lib/core-classes/DecodersService.ts
```diff
@@ -53,12 +53,15 @@ export class DecodersService {
     this.decoders.set(decoder.deviceModel, decoder);
   }
 
-  registerPayloadController(api: PluginApiDefinition, payloadService: PayloadService) {
+  getPayloadController(payloadService: PayloadService) {
+    const controllers = { actions: {} };
```
Aschen, line 57: Here you can use the `ControllerDefinition` type for this variable. (Also nitpicking, the name is not very clear to me)

Aschen (approved):

PR #133, opened 2021-11-09
typo & add type: auto for provisioning

rolljee:
# Description

This PR change the `autoProvisioning` property and standardize it with the existing catalog one.

### DIFF
--- a/lib/core-classes/PayloadService.ts
+++ b/lib/core-classes/PayloadService.ts
@@ -117,7 +117,7 @@ export class PayloadService {
   /**
    * Device provisioning strategy.
    *
-   * If autoProvisionning is on, device is automatically registered, otherwise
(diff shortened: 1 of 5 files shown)


lib/core-classes/PayloadService.ts
```diff
@@ -117,7 +117,7 @@ export class PayloadService {
   /**
    * Device provisioning strategy.
    *
-   * If autoProvisionning is on, device is automatically registered, otherwise
+   * If provisioningStrategy is on auto, device is automatically registered, otherwise
```
Aschen, line 120:
```suggestion
   * If provisioningStrategy is "auto," device is automatically registered, otherwise
```

Aschen (approved):

PR #135, opened 2021-11-16
Allow the developper to import devices using csv a compatible syntax

### LINKED ISSUE #132: Provisioning via CSV import
Provisioning catalog import should be available from a CSV file

rolljee:
# Description

This PR allow the developer to import device within the app using the CSV format

## importDevices

Import devices from a CSV file

---

### HTTP

```http
URL: http://kuzzle:7512/_/device-manager/devices/_import[?refresh=wait_for]
Method: POST
```

### Other protocols

```js
{
  "controller": "device-manager/device",
  "action": "importDevices",
  "body": {
    "csv": "qos.battery,measures.temperature.degree\\n80,23.3"
  }
}
```

### Kourou

```bash
kourou device-manager/device:importDevices --body '{ "csv": "qos.battery,measures.temperature.degree\\n80,23.3" }'
```
---

## Body properties

- `csv`: a csv syntax compatible containing at least this two headers `tenantId,deviceId` with their corresponding values

### Optional:

- `refresh`: if set to `wait_for`, Kuzzle will not respond until the documents are indexed

---

## Response

```js
{
  "status": 200,
  "error": null,
  "index": "<index>",
  "controller": "device-manager/device",
  "action": "importDevices",
  "requestId": "<unique request identifier>",
  "result": {}
}
```

closes #132

### DIFF
--- a/features/DeviceController.feature
+++ b/features/DeviceController.feature
@@ -131,13 +131,13 @@ Feature: Device Manager device controller
       | measures.temperature.degree             | 23.3                               |
       | measures.temperature.origin.qos.battery | 80                                 |
 
-    Scenario: Link the same device to another asset should fail
+  Scenario: Link the same device to another asset should fail
     When I successfully execute the action "device-manager/device":"linkAsset" with args:
       | _id     | "DummyTemp-attached_ayse_unlinked" |
       | assetId | "PERFO-unlinked"                   |
     And I execute the action "device-manager/device":"linkAsset" with args:
       | _id     | "DummyTemp-attached_ayse_unlinked" |
-      | assetId | "TIKO-unlinked"                   |
+      | assetId | "TIKO-unlinked"                    |
     Then I should receive an error matching:
       | message | "Device \"DummyTemp-attached_ayse_unlinked\" is already linked to the asset \"PERFO-unlinked\" you need to detach it first." |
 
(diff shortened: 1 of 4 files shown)


rolljee:
missing documentation, incoming.

features/DeviceController.feature
```diff
@@ -352,3 +352,10 @@ Feature: Device Manager device controller
       | collection | "payloads"       |
     Then I should receive a result matching:
       | total | 1 |
+
+  Scenario: Import devices using csv
+    Given I successfully execute the action "device-manager/device":"importDevices" with args:
```
Aschen, line 357: I don't understand how importing the string `qos.battery,measures.temperature.degree\\n80,23.3` can create the device `DummyTemp-attached_ayse_unlinked` (which already exists and so an error should be thrown instead here)

lib/controllers/DeviceController.ts
```diff
@@ -248,6 +252,24 @@ export class DeviceController extends CRUDController {
       { query: { bool: { filter } } });
   }
 
+  async importDevices (request: KuzzleRequest) {
+    const body = request.getBody();
```
Aschen, line 256: Instead of retrieving the body and manually check for the `csv` string property presence, you can directly use the `getBodyString` method. In that way, a standard error message will be returned if the string does not exists

lib/core-classes/DeviceService.ts
```diff
@@ -373,6 +375,35 @@ export class DeviceService {
     return results;
   }
 
+  async importDevices(devices: JSONObject,
+    { strict, options }: { strict?: boolean, options?: JSONObject }) {
+      const results = {
```
Aschen, line 380: Usually we indent like this when arguments does not fit into a single line
```suggestion
  async importDevices(
    devices: JSONObject,
    { strict, options }: { strict?: boolean, options?: JSONObject },
  ) {
      const results = {
```

Aschen (approved):
Also, the title should be modified

Aschen (approved):

PR #147, opened 2021-12-13
Fix BatchController update method

xbill82:
Fix the `errors.find` callback in BatchController::update so that update errors are properly thrown.

`mUpdate` errors have the following shape
```
{
  document: {
    _id: 'foo',
    ...
  },
  reason: 'bla bla I complain about sth LOL'
}
```
The `errors.find` routine was looking for the `_id` field right at the root of the error object, whereas it is within the `document` object. The result was that the error was not thrown and `undefined` was returned by the function, since `successes[idx]` pointed to nothing. This has been fixed.

## TODO
Tests should be written for the BatchWriter, on the following cases
* Adding a field to a device/asset with `strict` mapping (meaningful error should be thrown)
* Creating an asset in a collection that doesn't exist (meaningful error should be thrown) - case of broken fixtures

These cases are theoretically fixed by this PR.


Aschen (approved):

Aschen (approved):

Njuelle (approved):

Aschen:
I will need to re-open a new PR on a different branch starting from master to release only this hotfix

Aschen:
See https://github.com/kuzzleio/kuzzle-device-manager/pull/148

PR #149, opened 2021-12-15
add provisioning event before and after

rolljee:
## Events

Two events when the provisioning happen, allowing to modify the device/admin catalog and/or the tenant catalog before and after it is provisioned:

```js
app.pipe.register('device-manager:device:provisioning:before', async ({ device, adminCatalog, tenantCatalog }) => {
  app.log.debug('before provisioning trigered');

  set(device, '_source.metadata.enrichedByBeforeProvisioning', true);

  return { device, adminCatalog, tenantCatalog };
})


app.pipe.register('device-manager:device:provisioning:after', async ({ device, adminCatalog, tenantCatalog }) => {
  app.log.debug('after provisioning trigered');

  return { device, adminCatalog, tenantCatalog };
})
```

### DIFF
--- a/features/fixtures/application/app.ts
+++ b/features/fixtures/application/app.ts
@@ -95,8 +95,6 @@ app.pipe.register('device-manager:device:provisioning:after', async ({ device, a
       device._id,
       device._source,
     );
-
-    console.log('device response', response);
   }
 
   return { device, adminCatalog, tenantCatalog };
--- a/features/fixtures/application/app.ts
+++ b/features/fixtures/application/app.ts
@@ -86,6 +86,19 @@ app.pipe.register('device-manager:device:provisioning:before', async ({ device, 
 app.pipe.register('device-manager:device:provisioning:after', async ({ device, adminCatalog, tenantCatalog }) => {
   app.log.debug('after provisioning trigered');
 
+  if (device._source.metadata.enrichedByBeforeProvisioning) {
+    set(device, '_source.metadata.enrichedByAfterProvisioning', true);
+
+    const response = await app.sdk.document.update(
+      'device-manager',
+      'devices',
+      device._id,
+      device._source,
+    );
+
+    console.log('device response', response);
+  }
+
   return { device, adminCatalog, tenantCatalog };
 })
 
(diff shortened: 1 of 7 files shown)


lib/core-classes/PayloadService.ts
```diff
@@ -151,45 +151,61 @@ export class PayloadService {
       throw new BadRequestError(`Device ${device._id} is not allowed for registration.`);
     }
 
-    const ret = await this.register(device, decoder, request, { refresh });
+    const hasContent = catalogEntry && catalogEntry.content && catalogEntry.content.tenantId;
```
Aschen, line 154: Nitpicking: IMHO naming is unclear, this variable is true if a tenant is defined in the catalog entry

lib/core-classes/PayloadService.ts
```diff
@@ -151,45 +151,61 @@ export class PayloadService {
...
+    const tenantCatalogEntry = hasContent ? await this.getCatalogEntry(catalogEntry.content.tenantId, device._id) : undefined;
+
+    const response = await global.app.trigger('device-manager:device:provisioning:before', {
+      device,
+      adminCatalog: catalogEntry,
+      tenantCatalog: tenantCatalogEntry,
+    });
+
+    const enrichedDevice = response.device ? response.device : device;
```
Aschen, line 163: If the pipe does not returns and object with the same content it's normal if it crashes. In the core we don't use this kind of silent substitution

Aschen (changes requested):

Aschen:
Also, can you add a functional test on the after event please?

features/fixtures/application/app.ts
```diff
@@ -74,6 +74,34 @@ app.pipe.register(`tenant:tenant-ayse:asset:measures:new`, async (request: Kuzzl
...
+    set(device, '_source.metadata.enrichedByAfterProvisioning', true);
+
+    const response = await app.sdk.document.update(
+      'device-manager',
+      'devices',
+      device._id,
+      device._source,
+    );
+
+    console.log('device response', response);
```
Aschen, line 99: Debug leftover

lib/core-classes/DeviceManagerEngine.ts
```diff
@@ -1,4 +1,4 @@
-import { Plugin } from 'kuzzle';
+import { EmbeddedSDK, Plugin, PluginContext } from 'kuzzle';
```
Aschen, line 1: Useless imports
```suggestion
import { Plugin } from 'kuzzle';
```

Aschen (approved):

PR #150, opened 2021-12-16
event before after attachtenant

rolljee:
## Events

Two events when this action is called, allowing to modify the device before it is attached to tenant:

```js
app.pipe.register('device-manager:device:attach-tenant:before', async ({ index, device }) => {
  app.log.debug('before attach-tenant trigered');
  set(device, 'body.metadata.enrichedByBeforeAttachTenant', true);
  return { index, device };
})
app.pipe.register('device-manager:device:attach-tenant:after', async ({ index, device }) => {
  app.log.debug('after attach-tenant trigered');
  return { index, device };
})
```

### DIFF
--- a/lib/core-classes/DeviceService.ts
+++ b/lib/core-classes/DeviceService.ts
@@ -85,9 +85,10 @@ export class DeviceService {
       const enrichedDocuments = [];
 
       for (const deviceDocument of deviceDocuments) {
-        const response = await global.app.trigger(
-          'device-manager:device:attach-tenant:before',
-          { index: document.tenantId, device: deviceDocument }
+        const response = await global.app.trigger('device-manager:device:attach-tenant:before', {
+            index: document.tenantId,
+            device: deviceDocument
+          }
         );
 
         enrichedDocuments.push(response.device);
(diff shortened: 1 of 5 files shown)


lib/core-classes/DeviceService.ts
```diff
@@ -86,9 +86,20 @@ export class DeviceService {
...
       const deviceDocuments = this.formatDevicesContent(devices, document);
+      const enrichedDocuments = [];
+
+      for (const deviceDocument of deviceDocuments) {
+        const response = await global.app.trigger(
+          'device-manager:device:attach-tenant:before',
+          { index: document.tenantId, device: deviceDocument }
+        );
+        
+        const deviceData = response && response.device ? response.device : deviceDocument;
```
Aschen, line 97: Same remark as on the other PR, it's the developer responsibility to return a correct payload in the pipe

lib/core-classes/DeviceService.ts
```diff
@@ -110,6 +121,13 @@ export class DeviceService {
 
       results.successes.concat(successes);
       results.errors.concat(errors);
+
+      for (const deviceDocument of enrichedDocuments) {
+        await global.app.trigger(
+          'device-manager:device:attach-tenant:before',
```
Aschen, line 127:
```suggestion
          'device-manager:device:attach-tenant:after',
```

Aschen: That's why we should add functional tests for every feature :sweat_smile: You could create a document with an ID computed from the device ID in a collection in a pipe on this event for example, and then just check if this document exists

Aschen (changes requested):
Can you add a functional test for the after event please?

Aschen (approved):

PR #153, opened 2021-12-20
[events] Add before/after events on asset:update

rolljee:
## Events

Two events are triggered by this action, that can be used as follow:

```js
app.pipe.register('device-manager:asset:update:before', async ({ asset, updates }) => {
  app.log.debug('before asset update triggered');

  set(updates, 'metadata.enrichedByBeforeAssetUpdate', true);

  return { asset, updates };
})

app.pipe.register('device-manager:asset:update:after', async ({ asset, updates }) => {
  app.log.debug('after asset update triggered');

  if (updates.metadata.enrichedByBeforeAssetUpdate) {
    set(updates, 'metadata.enrichedByAfterAssetUpdate', true);

    await app.sdk.document.update(
      updates.metadata.index,
      'assets',
      asset._id,
      updates,
    )
  }

  return { asset, updates };
})
```

### DIFF
--- a/lib/controllers/AssetController.ts
+++ b/lib/controllers/AssetController.ts
@@ -72,8 +72,7 @@ export class AssetController extends CRUDController {
     request.input.body = response.updates;
     const result = await super.update(request);
 
-    await global.app.trigger(
-      'device-manager:asset:update:after', {
+    await global.app.trigger('device-manager:asset:update:after', {
       asset,
       updates: result._source,
     });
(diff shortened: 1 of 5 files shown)


lib/controllers/AssetController.ts
```diff
@@ -53,6 +53,34 @@ export class AssetController extends CRUDController {
...
+    });
+
+    request.input.body = response.updates;
+    const result = await super.update(request);
+
+    await global.app.trigger(
+      'device-manager:asset:update:after', {
+      asset,
+      updates: result._source,
+    });
```
Aschen, line 79: Nitpicking
```suggestion
    await global.app.trigger('device-manager:asset:update:after', {
      asset,
      updates: result._source,
    });
```

Aschen (approved):

PR #170, opened 2022-04-15
Add API action te retrieve measures linked to an asset

Aschen:
## Description

When a measure is made, the device may be linked to an asset. We might want to find an asset measure history across many devices so we need to keep a property containing the `assetId`.

This PR save the `assetId` in the measure document and expose a new API action to retrieve measures history for an asset: `device-manager/asset:measures`

### DIFF
--- a/features/support/hooks.js
+++ b/features/support/hooks.js
@@ -81,12 +81,18 @@ Before({ timeout: 30 * 1000 }, async function () {
     truncateCollection(this.sdk, 'tests', 'events'),
   ]);
 
-  await this.sdk.query({
-    controller: 'admin',
-    action: 'loadFixtures',
-    refresh: 'false',
-    body: defaultFixtures,
-  });
+  try {
+    await this.sdk.query({
+      controller: 'admin',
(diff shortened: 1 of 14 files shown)


features/support/hooks.js
```diff
@@ -81,12 +81,18 @@ Before({ timeout: 30 * 1000 }, async function () {
...
+  try {
+    await this.sdk.query({
+      controller: 'admin',
+      action: 'loadFixtures',
+      refresh: 'false',
+      body: defaultFixtures,
+    });
+  }
+  catch (error) {
+    console.dir(error, { depth: 10 });
```
clementbolin, line 93: is it normal to use a `console.dir`here ? is there no `debug` function like in Kuzzle ?

clementbolin (changes requested):

features/support/hooks.js
Aschen: No since it's in the functional test framework and I want to display the entire error content

clementbolin (approved):

PR #176, opened 2022-04-27
add capacities to override some method of the plugin

TisserantG:
…de in some case

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-device-manager",
-  "version": "0.4.14",
+  "version": "0.4.16",
   "description": "Manage your IoT devices and assets. Choose a provisionning strategy, receive and decode payload, handle your IoT business logic.",
   "author": "The Kuzzle Team (support@kuzzle.io)",
   "repository": {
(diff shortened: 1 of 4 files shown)


Aschen:
Can you add a more explicit title for the PR please?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-device-manager",
-  "version": "0.4.14",
+  "version": "0.4.16",
```
Aschen, line 3: This should be done automatically by the release tool

Aschen (changes requested):

TisserantG:
> Can you add a more explicit title and proper label for the PR please?
Done

package.json
TisserantG: done

package.json
```diff
@@ -25,29 +25,30 @@
   "license": "Apache-2.0",
   "dependencies": {
     "csvtojson": "~2.0.10",
+    "kuzzle": "^2.18.0",
```
Aschen, line 28: kuzzle should be left as peerDependency only
```suggestion
```

Aschen (approved):

Issue #179, opened 2022-04-27
Deleting an asset with a device doesn't unlink device


jpdalbi:
When deleting an asset in an engine, the document is deleted and the answer of the API is:
{
    "action": "delete",
    "controller": "device-manager/asset",
    "error": null,
    "index": "laposte",
    "node": "knode-garrulous-pelican-77744",
    "requestId": "7c720f17-b881-4302-b64f-5a7fed61c5b7",
    "result": "caisse-mobile_krone_0008",
    "status": 200,
    "volatile": null
}

However, if a device is linked to the asset it remains linked in the engine and in the device-manager index.
model:"AbeewayITSpot"
reference:"deveui001"
engineId:"laposte"
assetId:"caisse-mobile_krone_0008"

It causes a runtime error when a payload is pushed by API:
{
    "action": "abeeway-it-spot",
    "controller": "device-manager/payload",
    "error": {
        "code": 67239938,
        "id": "plugin.runtime.unexpected_error",
        "message": "Caught an unexpected plugin error: Cannot convert undefined or null to object\nThis is probably not a Kuzzle error, but a problem with a plugin implementation.",
        "props": [
            "Cannot convert undefined or null to object"
        ],

PR #181, opened 2022-05-02
unlink device when remove assets and http endpoint corrections

TisserantG:
Unlink device when remove assets : (see issue : https://github.com/kuzzleio/kuzzle-device-manager/issues/179 )
Http endpoint correction for many actions.
Change kuzzle dependency version to remove header http problem.


lib/controllers/AssetController.ts
```diff
@@ -131,7 +133,16 @@ export class AssetController extends CRUDController {
   }
 
   async delete (request: KuzzleRequest) {
+    
     request.input.args.index = request.getString('engineId');
+    const refresh = request.getRefresh();
+    const strict = request.getBoolean('strict');
+
+
```
OlivierCavadenti, line 141: blank line to remove

lib/core-classes/AssetService.ts
```diff
@@ -116,4 +116,21 @@ export class AssetService {
...
+        'term': {
+          'assetId': {
+            'value': assetId
+          }
+        }
+      }
+    } ;
+    
+    return await this.sdk.document.search('device-manager', 'devices', filter);
+
```
OlivierCavadenti, line 133: remove blank line

package.json
```diff
@@ -38,7 +38,7 @@
     "eslint": "^8.10.0",
     "eslint-friendly-formatter": "^4.0.1",
     "eslint-loader": "^4.0.2",
-    "kuzzle": "^2.17.6",
+    "kuzzle": "^2.18.0",
```
OlivierCavadenti, line 41: 2.18.1

features/AssetController.feature
```diff
@@ -17,6 +17,33 @@ Feature: Device Manager asset controller
       | _id      | "outils-PERFO-asset_01" |
     Then The document "engine-kuzzle":"assets":"outils-PERFO-asset_01" does not exists
 
+  Scenario: Delete a linked asset
+    When I successfully execute the action "device-manager/asset":"create" with args:
+      | engineId       | "engine-ayse" |
+      | body.type      | "outils"        |
+      | body.model     | "PERFO"         |
+      | body.reference | "asset_02"      |
+    Then The document "engine-ayse":"assets":"outils-PERFO-asset_02" exists
```
Aschen, line 26: This step is not useful, the previous step will throw an error if anything goes wrong because if the `successfully` keyword

features/AssetController.feature
```diff
@@ -17,6 +17,33 @@ Feature: Device Manager asset controller
...
+    When I successfully execute the action "device-manager/device":"linkAsset" with args:
+        | _id                            | "DummyTemp-attached_ayse_unlinked" |
+        | assetId                        | "outils-PERFO-asset_02"             |
+        | body.metadata.index  | "engine-ayse"         |
+    Then The document "engine-ayse":"devices":"DummyTemp-attached_ayse_unlinked" content match:
+      | assetId | "outils-PERFO-asset_02" |
+    And  The document "device-manager":"devices":"DummyTemp-attached_ayse_unlinked" content match:
+      | assetId | "outils-PERFO-asset_02" |
```
Aschen, line 34: Same for those steps because this scenario is not testing the asset linking. It should assume that this part is working since another scenario is ensuring this

lib/core-classes/AssetService.ts
```diff
@@ -109,11 +98,24 @@ export class AssetService {
...
+    const filter = {
+      'query': {
+        'term': {
+          'assetId': {
+            'value': assetId
+          }
+        }
+      }
+    } ;
+    return await this.sdk.document.search('device-manager', 'devices', filter);
```
Aschen, line 118: Useless await.

Also, you should use the Koncorde syntax when it's possible because it will be portable between storage backend in the futur

lib/core-classes/AssetService.ts
```diff
@@ -109,11 +98,24 @@ export class AssetService {
...
-
     if (isNaN(parsed)) {
       throw new BadRequestError(`"${name}" is not a valid ISO8601 date`);
     }
-
     return parsed;
   }
+
+
+  async getDeviceByAsset(assetId :string) {
```
Aschen, line 108: Maybe we could add a new field in the Asset document containing the linked devices IDs. It would allows to update them in a easiest way without having to search (and optionally refresh).

What do you think?

Aschen (changes requested):
I saw that you are removing a lot of blank lines, in my opinion having blank lines separating blocs of instruction from different contexts help the readability, like paragraphs and new lines helps article readability.

lib/core-classes/AssetService.ts
TisserantG: That would be useful for many different case, I will implement that.

lib/core-classes/AssetService.ts
```diff
@@ -42,23 +39,18 @@ export class AssetService {
         }
       }
     };
-
```
Aschen, line 45: Why did you removed all the single blank lines here? I feel that when the code is separated in "paragraphs", it's easier to read, what do you think?

lib/core-classes/DeviceService.ts
```diff
@@ -250,6 +250,11 @@ export class DeviceService {
     asset._source.measures = measures;
     device._source.assetId = linkRequest.assetId;
 
+    if(!Array.isArray(asset._source.deviceIds)) {
+      asset._source.deviceIds=[];
+    }
```
Aschen, line 255: Coding standards. Can you modify the linter config to match our standard on this projet? You need to copy those files
  - https://github.com/kuzzleio/kuzzle/blob/2-dev/.eslintrc.json
  - https://github.com/kuzzleio/kuzzle/blob/2-dev/.eslintc-ts.json
 
 ```suggestion
    if(! Array.isArray(asset._source.deviceIds)) {
      asset._source.deviceIds = [];
    }
```

File: lib/core-classes/DeviceService.ts [hunk 10]
```diff
@@ -310,6 +318,16 @@ export class DeviceService {
     });
     device._source.assetId = null;
 
+    if(Array.isArray(asset._source.deviceIds)) {
```
Aschen, line 321: You should modify the `BaseAsset` class to ensure that `deviceIds` is always an array as we do for `measures`

lib/core-classes/DeviceService.ts
```diff
@@ -310,6 +318,16 @@ export class DeviceService {
...
+      const filteredDeviceList = [];
+      for(const linkedDevice of asset._source.deviceIds) {
```
Aschen, line 323: Nitpicking: You can use [Array.splice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) instead

lib/core-classes/DeviceService.ts
```diff
@@ -334,7 +352,8 @@ export class DeviceService {
         engineId,
         'assets',
         asset._id,
-        { measures: response.asset._source.measures },
+        { deviceIds: response.asset._source.deviceIds,
+          measures: response.asset._source.measures },
```
Aschen, line 356: Coding standards
```suggestion
        { 
          deviceIds: response.asset._source.deviceIds,
          measures: response.asset._source.measures 
        },
```

lib/types/BaseAssetContent.ts
```diff
@@ -16,4 +16,6 @@ export interface BaseAssetContent extends KDocumentContent {
   measures?: MeasureContent[],
 
   metadata?: JSONObject,
+
+  deviceIds?: string[]
```
Aschen, line 20: This should always be set as an empty array at least

Aschen (changes requested):

lib/core-classes/AssetService.ts
TisserantG: When I need to clean the code for easier reading, I split it in more little function or I add comment. 
Keep blank line in a function is not a reflex that I have, but I understand the idea.

lib/core-classes/DeviceService.ts
TisserantG: I'm not sure that can help me in this situation.

Aschen: If I understand well, you want to remove the `deviceId` from the asset's device IDs
```
> a = ['a', 'b', 'c']
[ 'a', 'b', 'c' ]
> a.splice(a.indexOf('b'), 1)
[ 'b' ]
> a
[ 'a', 'c' ]
```

lib/types/BaseAssetContent.ts
TisserantG: Done

lib/core-classes/DeviceService.ts
TisserantG: done

lib/core-classes/DeviceService.ts
TisserantG: Ok, I have read function definition to quickly.
I will not do the change because it will cause merge problem with the next PR that have other objects in the list and can't use splice.

lib/core-classes/DeviceService.ts
TisserantG: done

lib/core-classes/AssetService.ts
Aschen: Usually comments inside functions should be avoided because the naming choices should be self-explanatory.  
I do agree for the small function splitting but here the blank lines where dividing code blocks or instructions, for example
image

What bother you with extra blank lines?

features/AssetController.feature
TisserantG: done

features/AssetController.feature
TisserantG: done

lib/core-classes/AssetService.ts
TisserantG: The two blank line given in example was deleted two quickly, sorry :s
No particular problem with extra blank line (on one of my other job, the rule was "never extra blank line" and that have perhaps format me to delete all of them. )

lib/core-classes/AssetService.ts
TisserantG: this search is not useful in the current version and is deleted.

Aschen (approved):

PR #186, opened 2022-05-18
2-dev need to be up to date from 2-stable

TisserantG:

  - Step 1 :
  - Step 2 :
  - Step 3 :


Aschen (commented):
This PR only contains previously validated changes right?

TisserantG:
> 
Yes, the goal is only to synchronize the two branches (2-dev was late).
Only "particular change" that can be reviewed is the merge conflict of the last commit

Aschen:
No problems then :+1: In this kind of situation you can directly merge the branches I think

PR #199, opened 2022-07-27
add enum objects

TisserantG:
## What does this PR do ?

 Add possibility to create metadata object type that can contain enum of possible value.

You can use functional tests. 
Or you can create metadata with type enum and an objectValueList, then attach it to an assetCategory, create an asset witht the category and a metadata value.


lib/core-classes/AssetCategoryService.ts
```diff
@@ -88,11 +89,27 @@ export class AssetCategoryService {
...
+      if (! metadata.valueList.includes(value)) {
+        throw global.app.errors.get('device-manager', 'asset_controller', 'enum_metadata', metadata.name, value);
+      }
+    }
+    if (metadata.objectValueList) {
+      let find = false;
+      for (const objectValue of metadata.objectValueList) {
+        const fromattedObjectValue = this.formatMetadataForGet(objectValue.object);
+        if (isEqual(fromattedObjectValue, value)) {
+          find = true;
```
OlivierCavadenti, line 107: if find is true, it will never be false in the future, so you could break the for loop here

lib/mappings/metadataMappings.ts
```diff
@@ -8,12 +49,18 @@ export const metadataMappings = {
     name: {
       type: 'keyword'
     },
+    objectValueList: {
```
OlivierCavadenti, line 52: name suggestion: why not use plural instead of "List" ?
Like objectValues ? More concise

TisserantG: I think it's important to know that meadata will be saved as a list.
If I write MetadataValues, the naïve representation is : 
{
  color : "red",
  size : 55
  ...
}
So my opinion is "list" in the name is an important information.

lib/controllers/MetadataController.ts
```diff
@@ -40,4 +61,20 @@ export class MetadataController extends RelationalController {
...
+  async get (request: KuzzleRequest) {
+    const id = request.getId();
+    const engineId = request.getString('engineId');
+    const metadata = await this.sdk.document.get<MetadataContent>(engineId, this.collection, id);
+    const processedMetadata = JSON.parse(JSON.stringify(metadata._source)) as ProcessedMetadataContent;
+
+    if (metadata._source.objectValueList) {
+      processedMetadata.objectValueList = [];
+      for (const objectValue of metadata._source.objectValueList) {
+        processedMetadata.objectValueList.push(await this.assetCategoryService.formatMetadataForGet(objectValue.object));
```
Aschen, line 75: Those calls should be done in //

package.json
```diff
@@ -27,6 +27,7 @@
     "csvtojson": "~2.0.10",
     "json-stable-stringify": "^1.0.1",
     "kuzzle-plugin-commons": "https://github.com/kuzzleio/kuzzle-plugin-commons/archive/refs/tags/1.0.4.tar.gz",
+    "lodash.isequal": "^4.5.0",
```
Aschen, line 30: You can directly import the entirety of lodash since it's already included in Kuzzle dependencies it won't change anything.

Aschen (approved):
2 nitpickings and LGTM :+1:

PR #203, opened 2022-08-03
add default route for payload, and add engineId verification

TisserantG:
## What does this PR do ?

Add a default route for payload so KDM now can receive payload before decoder declaration.

send data to : 
http://localhost:7512/_/device-manager/payload/:devicename

### Other changes

add engineId verification on linkAsset endpoint
bugfix for metadata verification with objectenum type


lib/core-classes/DeviceService.ts
```diff
@@ -262,24 +262,29 @@ export class DeviceService {
...
-    const engineId = device._source.engineId;
+    const deviceEngineId = device._source.engineId;
 
-    if (! engineId) {
+    if (! deviceEngineId) {
       throw new BadRequestError(`Device "${device._id}" is not attached to an engine.`);
     }
 
+    if (deviceEngineId !== engineId) {
+      throw new BadRequestError(`Device "${device._id}" is not attached to given engine.`);
```
OlivierCavadenti, line 279: Need to update the error message no ?

OlivierCavadenti (approved):

PR #211, opened 2022-08-29
Improve DX for Decoder class

Aschen:
## What does this PR do ?

This PR implements the following DX changes for the Decoder class:
 - The constructor does not take any arguments
 - The device model is inferred from the class name (e.g. `AbeewayGPSDecoder` => device model name is `AbeewayGPS`) or use the `deviceModel` property
 - Measures declaration is done in the `measures` property as an array of objects containing a `name` and `type` properties
 - `DecodedPayload` is now a class with a `addMeasurement` method and should be returned from the `Decoder.decode` method

Full example:
```js
class DummyTempDecoder extends Decoder {
  constructor () {
    super();

    this.measures = [
      { name: 'mainBattery', type: 'battery' },
      { name: 'temperature', type: 'temperature' },
    ];

    this.payloadsMappings = {
      deviceEUI: { type: 'keyword' }
    };
  }

  async decode (payload: JSONObject): Promise<DecodedPayload<Decoder>> {
    const decodedPayload = new DecodedPayload<DummyTempDecoder>();

    decodedPayload.addMeasurement<TemperatureMeasurement>(
      payload.deviceEUI,
      'temperature',
      {
        measuredAt: Date.now(),
        type: 'temperature',
        values: {
          temperature: payload.register55,
        },
      });

    decodedPayload.addMeasurement<BatteryMeasurement>(
      payload.deviceEUI,
      'mainBattery',
      {
        measuredAt: Date.now(),
        type: 'battery',
        values: {
          battery: payload.batteryLevel * 100,
        },
      });

    return decodedPayload;
  }
}
```

### Other changes

The `Measurement` type now accepts an optional parameters who will be used to type the `measurement.values`:

```
export type TemperatureMeasurement = Measurement<{
  temperature: number;
}>;

const temperature: TemperatureMeasurement = {
  measuredAt: Date.now(),
  type: 'temperature',
  values: {
    // Trying to put another value here will raise a type error
    temperature: devicePayload.temperature,
  },
};
```

### DIFF
--- a/lib/types/DecodedPayload.ts
+++ b/lib/types/DecodedPayload.ts
@@ -7,29 +7,56 @@ import { Measurement } from './measures/MeasureContent';
  *
  * @example
  * const decodedPayload: DecodedPayload = {
- *   'BZH42AZF': [
- *     {
- *       deviceMeasureName: 'battery',
- *       measuredAt: 1655379939496,
- *       type: 'battery',
- *       values: { battery: 32 },
- *     },
- *   ],
- *   'IS7L8HK': [
- *     {
- *       deviceMeasureName: 'internal_temperature',
- *       measuredAt: 1655379939496,
- *       type: 'temperature',
- *       values: { temperature: -3 },
- *     },
- *     {
- *       deviceMeasureName: 'external_temperature',
- *       measuredAt: 1655379939496,
- *       type: 'temperature',
- *       values: { temperature: 39 },
- *     },
- *   ],
+ *   devices: {
+ *     'BZH42AZF': [
+ *       {
+ *         deviceMeasureName: 'battery',
+ *         measuredAt: 1655379939496,
+ *         type: 'battery',
+ *         values: { battery: 32 },
+ *       },
+ *     ],
+ *     'IS7L8HK': [
+ *       {
+ *         deviceMeasureName: 'internal_temperature',
+ *         measuredAt: 1655379939496,
+ *         type: 'temperature',
+ *         values: { temperature: -3 },
+ *       },
+ *       {
+ *         deviceMeasureName: 'external_temperature',
+ *         measuredAt: 1655379939496,
+ *         type: 'temperature',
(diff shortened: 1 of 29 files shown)


lib/types/DecodedPayload.ts
```diff
@@ -1,35 +1,57 @@
...
+  }
+
+  /**
+   * Add a new measure into the decoded payload
+   *
+   * @param deviceReference Device reference
+   * @param measureName Name of the decoded measure
+   * @param measurement Measurement object
+   */
+  addMeasurement<TMeasurement extends Measurement = Measurement> (
```
TisserantG, line 33: Does  generic is really useful here?

TisserantG (approved):

lib/types/DecodedPayload.ts
Aschen: I added it so when you want to add a measurement, the TS types will ensure that:
 - the name of the measure exists
 - the values gathered into the measures are correct

TisserantG: After reading again the example, I understand. 
When I code, I (most of the time) do :

```
    measurement: BatteryMeasurement =       {
        measuredAt: Date.now(),
        type: 'battery',
        values: {
          battery: payload.batteryLevel * 100,
        };
    decodedPayload.addMeasurement(
      payload.deviceEUI,
      'mainBattery',
       measurement
      });
```

so I do not need the generic type in this case.

But in a more concise TS code style, I'm agree that GerericType is usefull in this case.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-device-manager",
-  "version": "2.0.0-rc9",
+  "version": "2.0.0-rc18",
```
alexandrebouthinon, line 3: You should update your branch to get the latest `rc-19` as base for your work

alexandrebouthinon (approved):

PR #214, opened 2022-09-02
hotfix : link an asset with metadata to an assetcategory

TisserantG:
## What does this PR do ?

hotfix : link an asset with metadata to an assetcategory result previously to an error. 
Hotfix resolve the problem.


OlivierCavadenti (approved):

Aschen (commented):
We should add a fonctional test to ensure this bug will not be present in the future

Aschen (approved):

PR #218, opened 2022-09-07
Fix a possible null exception in `MeasureService.findAssetMeasureName`

ChillPC:
## What does this PR do ?

In `MeasureService.findAssetMeasureName`, if the `measureLink` found is `null`, getting its `assetMeasureName` would result in an Exception.


lib/core-classes/MeasureService.ts
```diff
@@ -404,9 +404,14 @@ export class MeasureService {
...
+    // TOSEE Convert to assert ?
+    if (! deviceLink) {
+      throw new PluginImplementationError(`The device "${device._id}" is not linked to the asset "${asset._id}"`);
+    }
+
     const measureLink = deviceLink.measureNamesLinks.find(
       nameLink => nameLink.deviceMeasureName === deviceMeasureName);
 
-    return measureLink.assetMeasureName;
+    return measureLink ? measureLink.assetMeasureName : undefined;
```
OlivierCavadenti, line 415: Typescript return type is "string" but this function can return "undefined" now, change to "string?"

Aschen (changes requested):
You need to add a functional test for this bugfix

lib/core-classes/MeasureService.ts
ChillPC: `Array.prototype.find()` function return an `undefined` if it doesn't find anything. I added it for standardization.

Aschen (approved):

PR #221, opened 2022-09-16
quick fix

TisserantG:
Avoid error when config.adminIndex is not equal to 'device-manager'


Aschen (changes requested):
The package-lock.json file should be updated also

TisserantG:
> The package-lock.json file should be updated also

done

Aschen (approved):

PR #224, opened 2022-09-21
fix(AssetController): pass options to search query

### LINKED ISSUE #223: Query arguments are not forwarded from `assets:search` to `document:search`
Query arguments are not forwarded from `assets:search` to `document:search`. As it is, we cannot for example use the parameter `lang` to use Koncorde DSL in our search requests.

```js
  async search (request: KuzzleRequest) {
    const engineId = request.getString('engineId');
    const { searchBody } = request.getSearchParams();
    const category = this.getFieldPath(request, 'category', null, 'asset-category');

    const res = await this.sdk.document.search<BaseAssetContent>(engineId, this.collection, searchBody); // Should be forwarded here using request object
    for (const hit of res.hits) {
      const document = await this.esDocumentToFormatted<BaseAssetContent>(engineId, this.collection, hit, [category]);
      this.assetCategoryService.formatDocumentMetadata(document);
    }
    return res;

  }
```

sebtiz13:
## What does this PR do ?
This PR fix #223 issue

### DIFF
--- a/features/AssetController.feature
+++ b/features/AssetController.feature
@@ -31,10 +31,7 @@ Feature: DeviceManager asset controller
       | body      | {"query":{"equals":{"_id":"outils-PERFO-asset_01"}}} |
       | size      | "1" |
       | lang      | "koncorde" |
-    Then I should receive a "hits" array of objects matching:
-      | model           | "PERFO"         |
-      | metadata.foobar | 42              |
-      | metadata.index  | "engine-kuzzle" |
+    Then I debug "result"
     When I successfully execute the action "device-manager/asset":"delete" with args:
       | engineId | "engine-kuzzle"         |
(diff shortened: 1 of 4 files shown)


features/AssetController.feature
```diff
@@ -26,6 +26,15 @@ Feature: DeviceManager asset controller
       | model           | "PERFO"         |
       | metadata.foobar | 42              |
       | metadata.index  | "engine-kuzzle" |
+    When I successfully execute the action "device-manager/asset":"search" with args:
```
alexandrebouthinon, line 29: You should takea look at the Kuzzle Core functional tests that use `document:search`: https://github.com/kuzzleio/kuzzle/blob/b58fa9b4b133ca92811fc6d7102fa5672cd35400/features/DocumentController.feature#L45

PR #228, opened 2022-10-11
Feat/optin

TisserantG:


lib/DeviceManagerPlugin.ts
```diff
@@ -201,6 +204,40 @@ export class DeviceManagerPlugin extends Plugin {
...
+          for (const document of documents) {
+            if (document._source?.metadata) {
+              document._source.metadata =
+                this.assetCategoryService.formatMetadataForES(
+                  document._source.metadata
+                );
+            }
+          }
+        }
+        return documents;
```
Aschen, line 239: This part could be DRY'ied

lib/models/Device.ts
```diff
@@ -62,6 +61,6 @@ export class Device {
   }
 
   public updateMetadata(metadata: JSONObject) {
-    this._source.metadata = _.merge(this._source.metadata, metadata);
+    this._source.metadata = this._source.metadata.concat(metadata);
```
Aschen, line 64: This can lead to duplicate metadata

Aschen (changes requested):
You should have starts from the `2-dev` branch because it had the new module architecture

lib/core-classes/AssetCategoryService.ts
```diff
@@ -235,11 +239,17 @@ export class AssetCategoryService {
...
+      }
+      const formattedMetadata: JSONObject = {};
+      for (const metadata of assetMetadata) {
+        formattedMetadata[metadata.key] = this.getValue(metadata.value);
+      }
+      return formattedMetadata;
     }
-    return formattedMetadata;
+    return {};
   }
```
sebtiz13, line 253: Use fast return are more simplier to understand and most efficient
Source: https://szymonkrajewski.pl/why-should-you-return-early/

```ts
formatMetadataForGet(assetMetadata: FormattedMetadata[]) {
  if (!assetMetadata) {
    return {};
  }

  if (!Array.isArray(assetMetadata)) {
    return assetMetadata;
  }

  const formattedMetadata: JSONObject = {};
  for (const metadata of assetMetadata) {
    formattedMetadata[metadata.key] = this.getValue(metadata.value);
  }
  
  return formattedMetadata;
}
```

lib/types/DeviceContent.ts
```diff
@@ -37,3 +33,16 @@ export interface DeviceContent extends KDocumentContent {
...
   engineId?: string;
 }
+
+export interface DeviceContent extends BaseDeviceContent {
+  /**
+   * Device metadata
+   */
+  metadata?: JSONObject;
+}
+export interface EsDeviceContent extends BaseDeviceContent {
```
sebtiz13, line 43:
```suggestion
}

export interface EsDeviceContent extends BaseDeviceContent {
```

TisserantG:
> You should have starts from the `2-dev` branch because it had the new module architecture

Yes, but when I start to work, the new architecture was not merged :-(

Aschen:
> > You should have starts from the `2-dev` branch because it had the new module architecture
> 
> Yes, but when I start to work, the new architecture was not merged :-(

I checked and it was not that hard, I did it ;-)

lib/DeviceManagerPlugin.ts
TisserantG: I can do something like 

```
pipeFunction(myFunction)
        if (
          (request.input.args.collection === "assets" ||
            request.input.args.collection === "devices") &&
          request.input.args.options?.prettify
        ) {
          for (const document of documents) {
            if (document._source?.metadata) {
              document._source.metadata =
                myFunction(
                  document._source.metadata
                );
            }
          }
        }
...
this.pipes = {
     ....
      "generic:document:afterGet": async (documents, request: Request) => {
       return pipeFunction( this.assetCategoryService.formatMetadataForGet);}
      "generic:document:beforeUpdate": async (documents, request: Request) => {
       return pipeFunction(this.assetCategoryService.formatMetadataForES);}
};
```
It's more concise, but, in my opinion, (even with better naming) less readable.

lib/core-classes/AssetCategoryService.ts
TisserantG: I know. I got a very bad habit.

TisserantG: done

lib/models/Device.ts
TisserantG: revert

PR #230, opened 2022-10-25
correct format geo_point

sebtiz13:
When send metadata "geolocation" with `geo_point` like:
```json
{
  "geolocation": {
    "lat": 0,
    "lon": 0
  }
}
```

The data are formated in object
```json
{
  "key": "geolocation",
  "value": {
    "object": [
      {
        "key": "lat",
        "value": {
          "integer": 0
        }
      },
      {
        "key": "lon",
        "value": {
          "integer": 0
        }
      }
    ]
  }
}
```

Instead of geo_point like
```json
{
  "key": "geolocation",
  "value": {
    "geo_point": {
      "lat": 0,
      "lon": 0
    }
  }
}
```

It's due to this test:
https://github.com/kuzzleio/kuzzle-device-manager/blob/26d87060eb5a8855737a40f7e78582928b08802b/lib/modules/asset-category/AssetCategoryService.ts#L205

It's not "safe" because value `0` are considered like `falsy` so in this case format in geo_point are ignored and pass in the next condition "format like object" and produce bad metadata.
So In this case prefer test like `typeof value.lat === "undefined"` or test with hasOwnProperty `Object.prototype.hasOwnProperty.call(value, "lat")`

### DIFF
--- a/lib/modules/asset-category/AssetCategoryService.ts
+++ b/lib/modules/asset-category/AssetCategoryService.ts
@@ -202,10 +202,15 @@ export class AssetCategoryService {
       formattedValue.integer = value;
     } else if (typeof value === "boolean") {
       formattedValue.boolean = value;
-    } else if (value.lat) {
-      formattedValue.geo_point = value;
     } else if (typeof value === "object") {
-      formattedValue.object = this.formatMetadataForES(value);
+      if (
+        Object.prototype.hasOwnProperty.call(value, "lat") &&
+        Object.prototype.hasOwnProperty.call(value, "lon")
+      ) {
+        formattedValue.geo_point = value;
+      } else {
+        formattedValue.object = this.formatMetadataForES(value);
+      }
     } else {
       formattedValue.keyword = value;
     }
--- a/lib/modules/asset-category/AssetCategoryService.ts
+++ b/lib/modules/asset-category/AssetCategoryService.ts
@@ -204,8 +204,8 @@ export class AssetCategoryService {
       formattedValue.boolean = value;
     } else if (typeof value === "object") {
       if (
-        Object.prototype.hasOwnProperty.call(value, "lat") &&
-        Object.prototype.hasOwnProperty.call(value, "lon")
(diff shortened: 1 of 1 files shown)


lib/modules/asset-category/AssetCategoryService.ts
```diff
@@ -202,10 +202,15 @@ export class AssetCategoryService {
       formattedValue.integer = value;
     } else if (typeof value === "boolean") {
       formattedValue.boolean = value;
-    } else if (value.lat) {
-      formattedValue.geo_point = value;
     } else if (typeof value === "object") {
-      formattedValue.object = this.formatMetadataForES(value);
+      if (
+        Object.prototype.hasOwnProperty.call(value, "lat") &&
+        Object.prototype.hasOwnProperty.call(value, "lon")
```
Aschen, line 208: You could just test existence with `value.lat !== undefiner && value.lon !== undefined`

Aschen (approved):
You can release this as the rc27 while we are waiting for the refactor :+1:

Aschen:
@sebtiz13 Can you please:
 - remove the `fix(metadata)` from the PR title since the type of PR is already enforced by labels
 - add the `changelog-bug-fixes` label
 - assign yourself this PR

lib/modules/asset-category/AssetCategoryService.ts
sebtiz13: `hasOwnProperty` it's more safer because it test the property are directly in object not in prototype.
In most case I prefer this approach for more robust code, but in this case, it's sure just test undefined can be sufficient.

So just prefer test `typeof value.property === "undefined"`, it's more verbose but prevent re-declaration of `undefined`.

CF: https://javascript.plainenglish.io/the-different-ways-to-check-for-undefined-acaaa385a719

Aschen: Since prototype polution is forbidden as for redefining `undefined` I think the flaw described in this article does not applies to us

PR #232, opened 2022-10-25
[breaking] Prepare version 2.0 stable

Aschen:
## What does this PR do ?

Asset metadata will now be stored simply as objects. 
Assets only have a `model` and a `reference` as identifiers, like Devices.
```
{
  model: "container",
  reference: "ABC123",
  metadata: {
    weight: 200
  }
}
```

A lot of API action has been removed to keep the plugin as simple as possible.

 - refactor the entire code following the module separation

###  From https://github.com/kuzzleio/kuzzle-device-manager/pull/231 (BREAKING)

This PR add a `Model` module to manage 
 - Asset: declare models and register metadata for them
 - Device: register metadata for device models
 - Measure: declare new measure types

Models can be registered on the framework with one of the `plugin.models.registerXXXX` method or at runtime with the `ModelController`:
 - writeAsset|Device|Measure: upsert a model
 - deleteAsset|Device|Measure: delete a model
 - listAsset|Device|Measure: list models

Decoder are now registered alongside with a Device model
```
deviceManager.models.registerDevice("DummyTemp", new DummyTempDecoder(), {
  color: { type: "keyword" },
});
```

### From https://github.com/kuzzleio/kuzzle-device-manager/pull/234

Add default roles to manage plugin ressources. Admins roles can do everything and reader can only read.
 - `assets.admin`
 - `assets.reader`
 - `device.admin`: this role cannot attach and detach devices from engines
 - `device.reader`
 - `decoders.admin`
 - `measures.reader`
 - `measures.admin` 

 - device and asset models now have `defaultValues` for their metadata. Those values will be copied into assets and devices upon creation
 - fix bug that prevented metadata to be propagated and saved in measures collection

### From https://github.com/kuzzleio/kuzzle-device-manager/pull/235 (BREAKING)

 - measures document
 - remove measure unit
 - embedded measures
 - measure names
 - measures definition in digital twin models


Shiranuit (changes requested):
I still have about 50% to review

PR #241, opened 2022-12-14
Export roles as explicit variables

Aschen:
## What does this PR do ?

Export explicit variables containing roles informations and description.

E.g. `RoleDecodersAdmin`

```
/**
 * Role used to manage decoders and their payloads.
 *
 * It's a platform admin role.
 *
 * @example
    "device-manager/decoders": {
      actions: {
        "*": true,
      },
    },
 */
export const RoleDecodersAdmin: KuzzleRole = {
  name: 'decoders.admin',
  definition: {
    controllers: {
      "device-manager/decoders": {
        actions: {
          "*": true,
        },
      },
    },
  }
}
```


lib/core/DeviceManagerPlugin.ts
```diff
@@ -35,10 +35,11 @@ import { DeviceManagerEngine } from "./DeviceManagerEngine";
 import { DecodersRegister } from "./registers/DecodersRegister";
 import { InternalCollection } from "./InternalCollection";
 import { ModelsRegister } from "./registers/ModelsRegister";
+import { KuzzleRole } from "lib/modules/shared/types/KuzzleRole";
 
 export class DeviceManagerPlugin extends Plugin {
   public config: DeviceManagerConfiguration;
-  public roles = {};
+  public roles: KuzzleRole[] = [];
```
alexandrebouthinon, line 42: As you update the type of this class property (from object to array), should we need information about already deployed environments migration process? Maybe this is not useful, I don't realize the implications of this update

alexandrebouthinon (approved):

lib/core/DeviceManagerPlugin.ts
Aschen: This property is only used inside the plugin so there will be no impact

PR #252, opened 2023-01-06
Fix Asset and Device creation from backend

Aschen:
## What does this PR do ?

Every modification of a Digital Twin (Asset or Device) through associated controllers is contextualized with the user who made the request to keep a trace in Kuzzle metadata.

This allows for example to know who has modified a metadata or link a new device to an asset by consulting history documents in `assets-history` collection.

When a request is made to those controllers from the backend, there is no user since it's the EmbeddedSDK executing the request and it was causing an error.

This PR fix that, when the EmbeddedSDK is used to execute one of those action, it will use the user if the sdk was contextualized with [sdk.as](https://docs.kuzzle.io/core/2/guides/develop-on-kuzzle/embedded-sdk/#user-impersonation) method, otherwise the metadata will be set with the `null` user ID who represent the backend user.

### DIFF
--- a/lib/modules/device/DeviceService.ts
+++ b/lib/modules/device/DeviceService.ts
@@ -42,7 +42,13 @@ export class DeviceService {
   }
 
   private get impersonatedSdk() {
-    return (user: User) => this.sdk.as(user, { checkRights: false });
+    return (user: User) => {
+      if (user?._id) {
+        return this.sdk.as(user, { checkRights: false })
+      }
+
+      return this.sdk
+    };
   }
 
   constructor(plugin: Plugin) {
--- a/lib/modules/device/DeviceService.ts
+++ b/lib/modules/device/DeviceService.ts
@@ -44,10 +44,10 @@ export class DeviceService {
   private get impersonatedSdk() {
     return (user: User) => {
       if (user?._id) {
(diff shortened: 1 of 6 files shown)


lib/modules/device/DeviceService.ts
```diff
@@ -42,7 +42,13 @@ export class DeviceService {
...
   private get impersonatedSdk() {
-    return (user: User) => this.sdk.as(user, { checkRights: false });
+    return (user: User) => {
+      if (user?._id) {
+        return this.sdk.as(user, { checkRights: false });
+      }
+
+      return this.sdk;
+    };
   }
```
alexandrebouthinon, line 52: Since this is a very useful method should we provide it as well in the Framework?

alexandrebouthinon (approved):

lib/modules/device/DeviceService.ts
Aschen: Well actually Kuzzle could just return the normal SDK if the `user` is null but I'm not sure about potential side effects for users who relied on the fact that the method will throw if an incorrect user is passed.

Also, I don't know where we could put this method in the framework. Let's see if we need it another time and talk about this

PR #273, opened 2023-02-28
Fix addition of asset context unlinked measures

Aschen:
## What does this PR do ?

The asset context is added to each measure linked to an asset before it is saved in the measures collection.

A bug was adding the context to every measure of a device, even if they were not linked to an asset.

For example, a device with temperature and battery measure who was linked to an asset only with it's temperature measure would see it's battery measure with the asset context.

This was causing the action `device-manager/assets:getMeasures` to return an incorrect list of measures

### Other changes
 - migrate some tests to Jest


tests/modules/assets/action-get-measures.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  beforeEach(async () => {
+    await beforeEachTruncateCollections(sdk);
+    await beforeEachLoadFixtures(sdk);
+  });
+
+  afterAll(async () => {
+    sdk.disconnect();
+  });
+
+  it("should returns measures linked to the asset", async () => {
```
OlivierCavadenti, line 31: nit: split this test in three independant tests. You include all three in another describe function, with a beforeEach:

  ```
  await sendDummyTempPayloads(sdk, [
      { deviceEUI: "unlinked1", temperature: 12 },
      { deviceEUI: "unlinked1", temperature: 14 },
      { deviceEUI: "linked1", temperature: 20 },
      { deviceEUI: "linked1", temperature: 22 },
      { deviceEUI: "linked1", temperature: 24 },
      { deviceEUI: "linked2", temperature: 42 },
      { deviceEUI: "linked2", temperature: 44 },
    ]);
    await sdk.collection.refresh("engine-ayse", "measures");
```

File: tests/modules/ingestion-pipeline/propagation.test.ts [hunk 2]
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  beforeEach(async () => {
+    await beforeEachTruncateCollections(sdk);
+    await beforeEachLoadFixtures(sdk);
+  });
+
+  afterAll(async () => {
+    sdk.disconnect();
+  });
```
OlivierCavadenti, line 25: You can encapsulate all that code above in a single function and call it each time your need:

```js
export function initSdk(): Kuzzle {
  const sdk = useSdk();

  beforeAll(async () => {
    await sdk.connect();
    await beforeAllCreateEngines(sdk);
  });

  beforeEach(async () => {
    await beforeEachTruncateCollections(sdk);
    await beforeEachLoadFixtures(sdk);
  });

  afterAll(async () => {
    sdk.disconnect();
  }); 

return sdk;
}
```

then use like:

`const sdk = initSdk();`

tests/modules/assets/action-get-measures.test.ts
Aschen: The 3 things I'm testing are actually related to the same context, I just need to check 3 documents. I used the extra blocks because otherwise I cannot redeclare the `result` variable

OlivierCavadenti (approved):

PR #274, opened 2023-03-03
Convert Cucumber tests to Jest

Aschen:
## What does this PR do ?

I came up with a great idea of using Cucumber steps to write the code for Jest test cases and I actually did it but by mistake I removed the code of the generator after I successfully generated the code for every test :man_facepalming: 

Anyway, it's not very difficult: 
 - add a class `JestConverter` with write necessary strings (beginning of test file, beginning of test case, end of test case, etc) and store it in the `global` object
 - move all existing steps
 - run cucumber and declare all the steps needed
   - the steps should write the code into the file instead of actually testing something

If needed, I will do it again later. 

EDIT: I did it again and kept the migration framework here https://gist.github.com/Aschen/b8615c69d01cda8f86725a8944193d01


package.json
```diff
@@ -32,15 +30,13 @@
...
+    "axios": "^1.3.2",
     "ergol": "^1.0.2",
     "eslint-plugin-kuzzle": "^0.0.6",
-    "jest": "^29.4.3",
-    "kuzzle": "^2.21.0",
-    "kuzzle-sdk": "^7.10.6",
-    "should": "^13.2.3",
+    "jest": "^29.4.1",
+    "kuzzle": "^2.20.2",
+    "kuzzle-sdk": "^7.10.5",
```
OlivierCavadenti, line 39: Why change versions here ?

tests/scenario/migrated/asset-controller.test.ts
```diff
@@ -0,0 +1,244 @@
...
+  });
+
+  beforeEach(async () => {
+    await beforeEachTruncateCollections(sdk);
+    await beforeEachLoadFixtures(sdk);
+  });
+
+  afterAll(async () => {
+    sdk.disconnect();
+  });
```
OlivierCavadenti, line 24: You can improve the generation framework by generate all of that part (generated by   writeFileBeginning ()) in one function inside a separated file imported in each generated file.
As I can see, it's not dependant of the test file and never change.

OlivierCavadenti (changes requested):
I think you can move all "migrated" jest test files in dedicated directories in "modules" directory ?
like "assets", "devices", or you want to keep the split between "migrated" vs handcrafted ?

package.json
Aschen: I bumped to the latest version mainly because I needed the latest sdk version

tests/scenario/migrated/asset-controller.test.ts
Aschen: Unfortunately I don't have the framework anymore to modify everything again and anyway those tests shouldn't move so I think we can leave it as is

PR #283, opened 2023-03-24
Store reason of discarded payloads

OlivierCavadenti:
- Store if a payload is skipped, valid or error.
- Store the reason of the error, when a user defined validation error occurs.

### DIFF
--- a/lib/modules/decoder/PayloadService.ts
+++ b/lib/modules/decoder/PayloadService.ts
@@ -86,7 +86,7 @@ export class PayloadService {
         valid,
         payload,
         apiAction,
-        state,
+        DecodingState[state],
         errorReason
       );
     }
@@ -151,7 +151,7 @@ export class PayloadService {
     valid: boolean,
     payload: JSONObject,
     apiAction: string,
-    state?: DecodingState,
+    state?: string,
     reason?: string
   ) {
     try {
--- a/lib/modules/decoder/PayloadService.ts
+++ b/lib/modules/decoder/PayloadService.ts
@@ -14,8 +14,8 @@ import { ask, onAsk } from "../shared";
 import { DecodedPayload } from "./DecodedPayload";
 import { Decoder } from "./Decoder";
 import { AskPayloadReceiveFormated } from "./types/PayloadEvents";
-import { StateDecoder } from './StateDecoder';
-import { SkipError } from './SkipError';
+import { StateDecoder } from "./StateDecoder";
+import { SkipError } from "./SkipError";
 
 export class PayloadService {
   private config: DeviceManagerConfiguration;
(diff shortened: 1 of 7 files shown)


lib/modules/decoder/PayloadService.ts
```diff
@@ -57,23 +59,36 @@ export class PayloadService {
 
     const uuid = request.input.args.uuid || uuidv4();
     let valid = true;
+    let state: StateDecoder = StateDecoder.VALID;
```
Aschen, line 62:
```suggestion
    let state = StateDecoder.VALID;
```

lib/modules/decoder/StateDecoder.ts
```diff
@@ -0,0 +1,5 @@
+export enum StateDecoder {
```
Aschen, line 1: Maybe `DecodingState` is a better name otherwise it's confusing with the Decoder class

lib/modules/decoder/PayloadService.ts
```diff
@@ -57,23 +59,36 @@ export class PayloadService {
...
+      // TODO: Temporary workaround to prevent breaking anything; in the future,
+      // consider modifying the return value of the 'validate' function to return an object
       if (!valid) {
-        return { valid };
+        throw new SkipError("Skip by user defined validation");
       }
     } catch (error) {
       valid = false;
+      errorReason =
+        error.message || (error.toJSON ? error.toJSON().message : undefined);
```
Aschen, line 76: Why using `toJSON` ? If it's an Error or a KuzzleErro then we always have the `message` property

lib/modules/decoder/SkipError.ts
```diff
@@ -0,0 +1,5 @@
+export class SkipError {
+  constructor(message: string) {
+    // needed for match Error constructor signature
```
Aschen, line 3: You should propagate the message so user may skip message with a reason also

Aschen (changes requested):
This feature should be documented here https://github.com/kuzzleio/kuzzle-device-manager/blob/master/doc/2/guides/measures-ingestion/index.md?plain=1#L109

And also add the `state` and `reason` here https://github.com/kuzzleio/kuzzle-device-manager/blob/master/doc/2/guides/measures-ingestion/index.md?plain=1#L232

OlivierCavadenti:
> [ingestion/index.md?plain=1#L109](https://github.com/kuzzleio/kuzzle-device-manager/blob/master/doc/2/guides/measures-ingestion/index.md?plain=1#L109)
> 
> And also add the `state` and `reason` here https://github.com/kuzzleio/kuzzle-device-manager/blob/master/doc/2/guides/measures-ingestion/index.md?plain=1#L232

I still need to do that.
Btw, the state enum is stored in ES with a number, maybe could be better, for queries and clarity, to store the name of the value and not the number (0,1, 2) ? @Aschen

Aschen:
> Btw, the state enum is stored in ES with a number, maybe could be better, for queries and clarity, to store the name of the value and not the number (0,1, 2) 

Yes it would be better to have string values :+1:

Aschen (approved):

PR #286, opened 2023-03-28
ModelService upsert -> createOrReplace

fmauNeko:
## What does this PR do ?
This PR changes the `document.upsert` calls in `ModelService` into `document.createOrReplace` calls, in order to match the behavior described in the documentation.

It also updates the related tests.


Aschen (changes requested):
You should add a test where you are removing a default value for a model.

Also, can you assign the PR to yourself and add appropriate headers please?

fmauNeko:
Done !

Aschen (approved):
:tada:

PR #287, opened 2023-03-28
Update metadata in existing assets when the model is modified

### LINKED ISSUE #277: Update metadata in existing assets when the model is modified
Updating an asset (or device) model metadata should automatically update corresponding assets:
 - add metadata: initialize with `null` or the default metadata value if any
 - delete metadata: remove the value

fmauNeko:
## What does this PR do ?
This PR implements the propagation of any metadata addition / deletion on an asset model to the linked assets.

Fixes #277

### DIFF
--- a/lib/modules/asset/AssetService.ts
+++ b/lib/modules/asset/AssetService.ts
@@ -11,7 +11,7 @@ import _ from "lodash";
 
 import { AskDeviceUnlinkAsset } from "../device";
 import { MeasureContent } from "../measure/";
-import { AskModelAssetGet } from "../model";
+import { AskModelAssetGet, AssetModelContent } from "../model";
 import {
   AskEngineList,
   DeviceManagerConfiguration,
@@ -75,82 +75,7 @@ export class AssetService {
   registerAskEvents() {
     onAsk<AskAssetRefreshModel>(
       "ask:device-manager:asset:refresh-model",
-      async ({ assetModel }) => {
-        const engines = await ask<AskEngineList>(
-          "ask:device-manager:engine:list",
-          {
-            group: assetModel.engineGroup,
-          }
-        );
-
-        const targets = engines.map((engine) => ({
-          collections: [InternalCollection.ASSETS],
-          index: engine.index,
-        }));
-
-        const assets = await this.sdk.query<
-          BaseRequest,
-          JSONObject // TODO: switch to DocumentSearchResult<AssetContent> once KHit<> has index and collection properties
-        >({
-          action: "search",
-          body: { query: { equals: { model: assetModel.asset.model } } },
-          controller: "document",
-          lang: "koncorde",
-          targets,
-        });
-
-        const modelMetadata = {};
-
-        for (const metadataName of Object.keys(
-          assetModel.asset.metadataMappings
-        )) {
-          const defaultMetadata =
-            assetModel.asset.defaultMetadata[metadataName];
-          modelMetadata[metadataName] = defaultMetadata ?? null;
-        }
-
-        const removedMetadata: string[] = [];
-
-        const updatedAssetsPerIndex: Record<string, KDocument<AssetContent>[]> =
-          assets.result.hits.reduce(
-            (
-              acc: Record<string, KDocument<AssetContent>[]>,
-              asset: JSONObject
-            ) => {
-              const assetMetadata = { ...asset._source.metadata };
-
-              for (const key of Object.keys(asset._source.metadata)) {
-                if (!(key in modelMetadata)) {
-                  removedMetadata.push(key);
-                  delete assetMetadata[key];
-                }
-              }
-
-              asset._source.metadata = {
-                ...modelMetadata,
-                ...assetMetadata,
-              };
-
-              acc[asset.index].push(asset as KDocument<AssetContent>);
-
-              return acc;
-            },
-            Object.fromEntries(
-              engines.map((engine) => [
-                engine.index,
-                [] as KDocument<AssetContent>[],
-              ])
-            )
-          );
-
-        for (const [index, updatedAssets] of Object.entries(
-          updatedAssetsPerIndex
-        )) {
-          await this.mReplace(null, index, updatedAssets, removedMetadata, {
-            refresh: "wait_for",
-          });
-        }
-      }
+      this.refreshModel
     );
   }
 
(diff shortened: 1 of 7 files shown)


tests/scenario/modules/models/asset-model-metadata-propagation.test.ts
```diff
@@ -0,0 +1,361 @@
...
+          width: { type: "integer" },
+        },
+        defaultValues: {
+          depth: 80,
+          width: 120,
+        },
+      },
+    });
+
+    await sdk.query<ApiAssetCreateRequest>({
```
OlivierCavadenti, line 61: put assets creation in fixtures file ?

lib/modules/model/ModelService.ts
```diff
@@ -90,7 +96,32 @@ export class ModelService {
...
+        action: "search",
+        body: { query: { equals: { model } } },
+        controller: "device-manager/assets",
+        engineId: engine.index,
+        lang: "koncorde",
+      });
+
+      await Promise.all(
+        assets.result.hits.map((asset) => {
+          ask<AskAssetRefreshModel>("ask:device-manager:asset:refresh-model", {
```
OlivierCavadenti, line 117: ask is not awaited :'( :'(

OlivierCavadenti (changes requested):

tests/scenario/modules/models/asset-model-metadata-propagation.test.ts
fmauNeko: There's some variations in the assets between tests

lib/modules/asset/AssetService.ts
```diff
@@ -52,6 +60,43 @@ export class AssetService {
...
+        for (const metadataName of Object.keys(
+          assetModel.asset.metadataMappings
+        )) {
+          modelMetadata[metadataName] = null;
+        }
+        for (const [metadataName, metadataValue] of Object.entries(
+          assetModel.asset.defaultMetadata
+        )) {
+          _.set(modelMetadata, metadataName, metadataValue);
+        }
```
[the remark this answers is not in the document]
fmauNeko: Done !

lib/modules/asset/AssetService.ts
```diff
@@ -67,34 +75,84 @@ export class AssetService {
...
+            const assetMetadata = Object.fromEntries(
+              Object.entries(asset._source.metadata).filter(([key]) => {
+                if (!(key in modelMetadata)) {
+                  removedMetadata.push(key);
+                  return false;
+                }
+
+                return true;
+              })
+            );
```
tdislay, line 130: If the `delete` keyword is ok (in this case it seems to be)
Could you use a for loop like this:
```ts
for (const key of Object.keys(asset._source.metadata) {
  if (!(key in modelMetadata)) {
    removedMetadata.push(key);
    delete asset._source.metadata[key];
  }
}
```
suggested by @sebtiz13

tdislay (approved):

lib/modules/asset/AssetService.ts
```diff
@@ -67,34 +75,84 @@ export class AssetService {
...
         }
 
-        const removedMetadata = _.difference(
-          Object.keys(asset._source.metadata),
-          Object.keys(modelMetadata)
+        const removedMetadata: string[] = [];
+
+        const updatedAssetsPerIndex: {
+          [x: string]: KDocument<AssetContent>[];
+        } = assetSearch.result.hits.reduce(
```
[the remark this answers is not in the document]
fmauNeko: Probably none.

lib/modules/asset/AssetService.ts
fmauNeko: It should indeed be okay

fmauNeko: It should indeed be okay !

lib/modules/asset/AssetService.ts
```diff
@@ -52,6 +68,90 @@ export class AssetService {
...
     this.assetHistoryService = assetHistoryService;
+
+    this.registerAskEvents();
+  }
+
+  registerAskEvents() {
+    onAsk<AskAssetRefreshModel>(
+      "ask:device-manager:asset:refresh-model",
+      async ({ assetModel }) => {
+        const engines = await ask<AskEngineList>(
```
Aschen, line 79: Can you declare a private method for clarity please?

lib/modules/asset/AssetService.ts
```diff
@@ -52,6 +68,90 @@ export class AssetService {
...
+                engine.index,
+                [] as KDocument<AssetContent>[],
+              ])
+            )
+          );
+
+        for (const [index, updatedAssets] of Object.entries(
+          updatedAssetsPerIndex
+        )) {
+          await this.mReplace(null, index, updatedAssets, removedMetadata, {
```
Aschen, line 149: You can parallelize those calls

lib/modules/asset/AssetService.ts
```diff
@@ -306,6 +406,45 @@ export class AssetService {
     return result;
   }
 
+  /**
+   * Replace an asset metadata
+   */
+  public async mReplace(
```
Aschen, line 412: Nit: from the behavior this function is replacing and historizing no?

Aschen (changes requested):

lib/modules/asset/AssetService.ts
fmauNeko: That's what it does indeed

lib/modules/asset/AssetService.ts
```diff
@@ -306,6 +406,45 @@ export class AssetService {
...
+    user: User,
+    engineId: string,
+    assets: KDocument<AssetContent>[],
+    removedMetadata: string[],
+    { refresh }: { refresh: any }
+  ): Promise<mReplaceResponse> {
+    const replacedAssets = await this.impersonatedSdk(
```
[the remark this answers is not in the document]
fmauNeko: `user` is fixed at `null` on the call in the `Ask`. I kept the `user` parameter for consistency with the other SCRUD methods.

lib/modules/asset/AssetService.ts
```diff
@@ -52,6 +68,15 @@ export class AssetService {
...
     this.config = plugin.config;
     this.assetHistoryService = assetHistoryService;
+
+    this.registerAskEvents();
+  }
+
+  registerAskEvents() {
+    onAsk<AskAssetRefreshModel>(
+      "ask:device-manager:asset:refresh-model",
+      this.refreshModel
```
Aschen, line 78: The context is not bound for `refreshModel` (`this` will not be available inside the method)

lib/modules/asset/AssetService.ts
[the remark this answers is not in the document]
Aschen: I understand but here there is no need to use the ImpersonatedSDK for the reasons I gave in my previous message. This one can not really be considered as a SCRUD method I think

tdislay (approved):

Aschen (approved):

PR #288, opened 2023-03-29
Full path nested metadata when create/update assets

### LINKED ISSUE #276: Add full path of modified nested metadata in history when updating an asset
When updating an asset nested metadata, the [complete path is not saved in the history document](https://github.com/kuzzleio/kuzzle-device-manager/blob/master/lib/modules/asset/AssetService.ts#L170) created.

This has been done when modifying an asset metadata inside the ingestion pipeline: https://github.com/kuzzleio/kuzzle-device-manager/blob/master/lib/modules/measure/MeasureService.ts#L215

It would be better to find a DRY solution

OlivierCavadenti:
- When an asset is created or updated, we saved the complete paths in asset history document.

Closes #276

### DIFF
--- a/lib/modules/asset/AssetService.ts
+++ b/lib/modules/asset/AssetService.ts
@@ -129,7 +129,7 @@ export class AssetService {
     engineId: string,
     assetId: string
   ): Promise<KDocument<AssetContent>> {
-    return await this.sdk.document.get<AssetContent>(
+    return this.sdk.document.get<AssetContent>(
       engineId,
       InternalCollection.ASSETS,
       assetId
--- a/lib/modules/asset/AssetService.ts
+++ b/lib/modules/asset/AssetService.ts
@@ -4,7 +4,7 @@ import { JSONObject, KDocument, KHit, SearchResult } from "kuzzle-sdk";
 
 import { MeasureContent } from "../measure/";
 import { AskDeviceUnlinkAsset } from "../device";
-import { EmbeddedMeasure, Metadata, lock, ask } from "../shared";
+import { EmbeddedMeasure, Metadata, lock, ask, flattenObject } from "../shared";
 import {
   DeviceManagerConfiguration,
   InternalCollection,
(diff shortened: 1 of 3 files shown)


lib/modules/asset/AssetService.ts
```diff
@@ -129,13 +129,11 @@ export class AssetService {
     engineId: string,
     assetId: string
   ): Promise<KDocument<AssetContent>> {
-    const asset = await this.sdk.document.get<AssetContent>(
+    return await this.sdk.document.get<AssetContent>(
```
Aschen, line 132: Useless await

tests/scenario/modules/assets/asset-history.test.ts
```diff
@@ -67,4 +70,129 @@ describe("DeviceController: receiveMeasure", () => {
...
+    });
+
+    await sdk.collection.refresh("engine-ayse", "assets");
+
+    const result2 = await sdk.document.get<AssetContent>(
+      "engine-ayse",
+      "assets",
+      "Container-linked1"
+    );
+    expect(result2._source.metadata).toMatchObject({
```
Aschen, line 120: You can use `document.get` directly in expect and avoid the `result2` variable

tests/scenario/modules/assets/asset-history.test.ts
```diff
@@ -67,4 +70,129 @@ describe("DeviceController: receiveMeasure", () => {
...
+    });
+
+    await sdk.collection.refresh("engine-ayse", "assets");
+
+    const result2 = await sdk.document.get<AssetContent>(
+      "engine-ayse",
+      "assets",
+      "Container-linked3"
+    );
+    expect(result2._source.metadata).toMatchObject({
```
Aschen, line 187: You can use `document.get` directly in expect and avoid the `result2` variable

Aschen (approved):

fmauNeko (approved):

PR #290, opened 2023-03-30
Historize asset for each same measurements

tdislay:
## What does this PR do ?

This PR fix an issue when multiple same measurements sent over one payload were ingested, only the most recent measure would be `historized`.
Also, it adds a `timestamp` property to assetHistoryContent. It allows you to sort documents by event emitted date rather that document `created_at`.

I also used a `mCreate` in the asset history add event instead of a simple `create`

### DIFF
--- a/lib/modules/asset/AssetService.ts
+++ b/lib/modules/asset/AssetService.ts
@@ -362,9 +362,7 @@ export class AssetService {
         asset: asset._source as AssetContent,
         event: {
           metadata: {
-            names: Object.keys(flattenObject(asset._source.metadata)).concat(
-              removedMetadata.map((name) => `-${name}`)
-            ),
+            names: Object.keys(flattenObject(asset._source.metadata)),
           },
           name: "metadata",
         },


lib/modules/measure/MeasureService.ts
```diff
@@ -106,7 +114,7 @@ export class MeasureService {
         asset,
         measurements,
         payloadUuids
-      );
+      ).sort((measureA, measureB) => measureA.measuredAt - measureB.measuredAt);
```
fmauNeko, line 117: Can you move the `sort` in the `buildMeasures` method ?

fmauNeko (changes requested):

lib/modules/measure/MeasureService.ts
```diff
@@ -237,6 +224,48 @@ export class MeasureService {
               })
           );
         }
+
+        for (const [measuredAt, assetState] of Object.entries(
```
Aschen, line 228: Can you extract the logic for asset historization into a method please?

Aschen (changes requested):

lib/modules/measure/MeasureService.ts
```diff
@@ -237,6 +224,48 @@ export class MeasureService {
...
+          const changes = objectDiff(
+            originalAssetMetadata,
+            asset._source.metadata
+          );
+
+          if (changes.length !== 0) {
+            (event as unknown as AssetHistoryEventMetadata).metadata = {
+              names: changes,
+            };
+          }
```
tdislay, line 255: I'm confused @Aschen @fmauNeko 
This code means, if a metadata was edited and that there's multiple assetStates, each asset-history document event will have a metadata property. Where, maybe, only one (the first / the last) should have.
It's no big deal, but I find it weird. What do you think ?

lib/modules/measure/MeasureService.ts
tdislay: done

lib/modules/measure/MeasureService.ts
tdislay: done

lib/modules/measure/MeasureService.ts
Aschen: Very good remark.

Actually if a business rule edit the asset metadata after some computation on a mesure then the metadata property should be set on the according asset-history document (the one containing the measure as well)

This would need to think about an framework API to bound a measure with a metadata edition.

I think for now you can just save the metadata with the last document, edit documentation accordingly to explain this behavior and add a comment for the futur

tdislay: There's currently no test for this behaviour. I presume I should write one ?

Aschen: Right your presume young padawan

tdislay: A test, I shall write

fmauNeko (approved):

lib/modules/measure/MeasureService.ts
```diff
@@ -300,10 +285,18 @@ export class MeasureService {
...
+  /**
+   * Update asset with each non-null non-computed measures.
+   *
+   * @returns An object of each asset state by timestamp.
+   */
   private updateAssetMeasures(
     asset: KDocument<AssetContent>,
     measurements: MeasureContent[]
-  ) {
+  ): Map<number, KDocument<AssetContent<any, any>>> {
```
OlivierCavadenti, line 297: cannot create a dedicated object instead of this ugly map ?

tests/application/decoders/DummyTempDecoder.ts
```diff
@@ -22,22 +22,43 @@ export class DummyTempDecoder extends Decoder {
...
+    return payloads.every((payload) => {
+      if (!payload.deviceEUI) {
+        throw new PreconditionError('Invalid payload: missing "deviceEUI"');
+      }
+
+      if (payload.invalid) {
+        return false;
+      }
+
+      return true;
```
OlivierCavadenti, line 41: => return ! payload.invalid

lib/modules/measure/MeasureService.ts
tdislay: I added [a comment](https://github.com/kuzzleio/kuzzle-device-manager/pull/290/commits/545c79a5bc3cdd0abb2613a965fccdd4e195fb51#diff-064fc947b9c2a22f50298738d3bf7f5cec0172e69d7c6abbd64b5c07409c5c1fR298) to indicate why we use a Map (preserve insertion order).

We can use a type alias but it wouldn't be used only in one file AND it wouldn't help clarity
I don't see better ways to do it:
```ts
type AssetStates = Map<number, KDocument<AssetContent<any,any>>
```
Moreover, we would have things like:
```ts
const assetStates: AssetStates = new Map<number, KDocument<AssetContent>>();
```

I'm confused, perhaps we should just leave it like this...

Ping @Aschen

tests/scenario/modules/assets/asset-history.test.ts
```diff
@@ -67,4 +136,64 @@ describe("DeviceController: receiveMeasure", () => {
...
+          },
+        ],
+      },
+    ]);
+    await sdk.collection.refresh("engine-ayse", "assets-history");
+
+    const result = await sdk.document.search<AssetHistoryContent>(
+      "engine-ayse",
+      "assets-history",
+      { sort: { timestamp: "desc" } }
```
Aschen, line 165: same comment about the sort

tests/scenario/modules/assets/asset-history.test.ts
```diff
@@ -35,6 +35,75 @@ describe("DeviceController: receiveMeasure", () => {
...
```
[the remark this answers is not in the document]
tdislay: Since asset historization promises are parallelized, they are not always created in the right order.
We added the property timestamp in order to retrieve those documents in the right order..

lib/modules/measure/MeasureService.ts
Aschen: Why not using an array of tuples? `Array<[measuredAt, asset]>`

tests/scenario/modules/assets/asset-history.test.ts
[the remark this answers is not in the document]
tdislay: Done

fmauNeko (approved):

lib/modules/measure/MeasureService.ts
```diff
@@ -444,3 +448,61 @@ export class MeasureService {
     return measureName.asset;
   }
 }
+
+/**
+ * Create a new document in the collection asset-history, for each asset states
+ */
+function historizeAssetStates(
```
Aschen, line 455: This method should be async so it will always return a promise instead of throw an exception.

Aschen (changes requested):

lib/modules/measure/MeasureService.ts
tdislay: Oupsy !
Done :)

lib/modules/asset/AssetHistoryService.ts
```diff
@@ -29,23 +28,33 @@ export class AssetHistoryService {
...
+
+    await this.sdk.document.mCreate<AssetHistoryContent<TAssetHistoryEvent>>(
       engineId,
       InternalCollection.ASSETS_HISTORY,
-      {
-        asset: asset._source,
-        event,
-        id: asset._id,
-      }
+      contents
```
Aschen, line 57: Explicitely throw if an error occur
```suggestion
      contents,
      { strict: true }
```

lib/modules/asset/AssetService.ts
```diff
@@ -348,21 +357,25 @@ export class AssetService {
...
-              ),
-            },
-            name: "metadata",
+    const histories: AssetHistoryContent[] = replacedAssets.successes.map(
+      (asset) => ({
+        asset: asset._source as AssetContent,
+        event: {
+          metadata: {
+            names: Object.keys(flattenObject(asset._source.metadata)).concat(
+              removedMetadata.map((name) => `-${name}`)
```
Aschen, line 366: The `names` array is supposed to contains the metadata names without additional information. To know if the metadata was removed/modified/added then users need to compare with previous history document

lib/modules/measure/MeasureService.ts
```diff
@@ -444,3 +448,61 @@ export class MeasureService {
...
+  assetStates: Map<number, KDocument<AssetContent<any, any>>>,
+  engineId: string,
+  originalAssetMetadata: Metadata,
+  assetMetadata: Metadata
+): Promise<void> {
+  const metadataChanges = objectDiff(originalAssetMetadata, assetMetadata);
+  const lastTimestampRecorded = Array.from(assetStates.keys()).pop();
+
+  const histories: AssetHistoryContent[] = [];
+  for (const [measuredAt, assetState] of assetStates) {
```
Aschen, line 465: Isn't it this?
```suggestion
  for (const [measuredAt, assetState] of assetStates.entries()) {
```

lib/modules/asset/AssetService.ts
tdislay: It seems to have been added recently by Florian, in order to fix [an issue](https://github.com/kuzzleio/kuzzle-device-manager/pull/287/files#diff-7afb8db4a98276c4e4429a694fb1610155bda81aa8a15cc17fdd093daaea4c13R356-R358)

lib/modules/measure/MeasureService.ts
tdislay: [Since `Map` is iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#iterating_map_with_for...of) you can do it both ways:
```ts
for (const [key, value] of myMap);
// AND
for (const [key, value] of myMap.entries());
```

Aschen: Oh ok I didn't knew. You can leave it like this :+1:

lib/modules/asset/AssetService.ts
Aschen: @fmauNeko Can you give us more insight on this please?

fmauNeko: I added it to clearly mark removed metadata. If that was wrong, feel free to revert it, or I can open a PR to remove that.

Aschen: The first goal was to be able to search for history document regarding a metadata to retrace modifications so for now it's better to keep the name only :+1: 

We could add more info with addition/deletion/modification if needed

PR #291, opened 2023-04-05
Add type of measure linked in each links of asset measures.

OlivierCavadenti:
- Add the type of the measure during link asset measure.

### DIFF
--- a/doc/2/controllers/devices/linkAsset/index.md
+++ b/doc/2/controllers/devices/linkAsset/index.md
@@ -33,8 +33,7 @@ Method: PUT
     "measureNames": [
       {
         "asset": "<name of the measure in the asset>",
-        "device": "<name of the measure in the device>",
-        "type": "<type of the measure in the device>"
+        "device": "<name of the measure in the device>"
       }
     ]
   },
--- a/doc/2/controllers/devices/linkAsset/index.md
+++ b/doc/2/controllers/devices/linkAsset/index.md
@@ -33,7 +33,8 @@ Method: PUT
     "measureNames": [
       {
         "asset": "<name of the measure in the asset>",
-        "device": "<name of the measure in the device>"
+        "device": "<name of the measure in the device>",
+        "type": "<type of the measure in the device>"
       }
     ]
   },
--- a/lib/modules/device/DeviceService.ts
+++ b/lib/modules/device/DeviceService.ts
@@ -469,7 +469,7 @@ export class DeviceService {
             `Measure "${measure.asset}" is not declared in the device model "${measure.device}".`
           );
         }
-        const type = findedMeasure.type;
+        const { type } = findedMeasure;
         updatedMeasureNames.push({
           asset: measure.asset,
           device: measure.device,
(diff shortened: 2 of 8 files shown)


doc/2/controllers/devices/linkAsset/index.md
```diff
@@ -33,7 +33,8 @@ Method: PUT
     "measureNames": [
       {
         "asset": "<name of the measure in the asset>",
-        "device": "<name of the measure in the device>"
+        "device": "<name of the measure in the device>",
+        "type": "<type of the measure in the device>"
```
Aschen, line 37: I think we don't need the type in the API, since we have the name then we can retrieve the type

Aschen (changes requested):

doc/2/controllers/devices/linkAsset/index.md
OlivierCavadenti: as I understand, I remove this from the doc and I need to update measureNames during linkingAssets by adding "type" from deviceModel.device.measures (and not request it in body request).

Aschen: Yes exactly :+1: The code is a little more complex but the API is easier for end users

lib/modules/device/DeviceService.ts
```diff
@@ -457,21 +459,33 @@ export class DeviceService {
         this.getDeviceModel(device._source.model),
       ]);
 
-      this.checkAlreadyProvidedMeasures(device, asset, measureNames);
+      let updatedMeasureNames: MeasureName[] = [];
+      for (const measure of measureNames) {
+        const type = deviceModel.device.measures.find(
+          (deviceMeasure) => deviceMeasure.name === measure.device
+        ).type;
```
Aschen, line 466: Are we sure that the provided measure name for the device exists in the device model definition?

OlivierCavadenti: I can check that before that. If it's null, I need to add the measure without type or I can find it in another place (on the asset and/or device) ?

Aschen: If the user want to link the asset and the device with a measure name that doesn't exists then we should throw an error (my bad I forgot this case I think)

OlivierCavadenti: In the case of implicit linking (in generateMissingAssetMeasureNames) it seems undeclared measures are not created. So if I throw an error, it could be a breaking change ?

Aschen: Yes you are right :+1: 

Then we should check only if implicit linking was not specified

Aschen (approved):

lib/modules/device/DeviceService.ts
```diff
@@ -457,21 +459,39 @@ export class DeviceService {
...
+      for (const measure of measureNames) {
+        let findedMeasure = deviceModel.device.measures.find(
+          (deviceMeasure) => deviceMeasure.name === measure.device
+        );
+        if (!findedMeasure && !implicitMeasuresLinking) {
+          throw new BadRequestError(
+            `Measure "${measure.asset}" is not declared in the device model "${measure.device}".`
+          );
+        }
+        const type = findedMeasure.type;
```
rolljee, line 472:
```suggestion
        const { type } = findedMeasure;
```

rolljee (approved):

Shiranuit (approved):

PR #302, opened 2023-05-02
fix: CSV column headers and path to values should not be the same

tdislay:
## What does this PR do ?

This PR fixes empty values when measure named differed from measure types.
Previously, when you had a measure named `temperatureExt` of type `temperature`, it tried to get the value with the current path : `values.temperatureExt` where it should have been `values.temperature`.

Plus, this allows us to use custom header names instead of the paths to the value.

### DIFF
--- a/lib/modules/measure/MeasureExporter.ts
+++ b/lib/modules/measure/MeasureExporter.ts
@@ -171,25 +171,29 @@ export class MeasureExporter {
         }
       );
 
-      const columns = [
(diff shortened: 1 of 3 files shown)


tdislay:
I used a Map in order to keep the key insertion order. Maybe, I should let a comment in the code about this

alexandrebouthinon (approved):

sebtiz13 (approved):

PR #307, opened 2023-06-01
chore: improve CI workflow

sebtiz13:
## What does this PR do ?

Update and improve CI workflows, refactor actions directly in workflows steps and add `check-types` step to check types error before try to build.

### Other changes

**Improve docker-compose**

- Open redis and elastic port for local development
- Add healthcheck on services to automatically restart when crash
- Use node 16 version of kuzzle container
- Add a volume to keep data of elasticsearch
- Add a volume for the `node_modules` of kuzzle backend (prevent error with packages installed in local, ex: GLIBC error)

### Boyscout

**Update kuzzle dependency**
Update `kuzzle` package from 2.23 to 2.25.
This is useful to update `uWebSockets.js` package from 20.0 to 20.7 to support node version 18


.github/workflows/push_dev.workflow.yml
```diff
@@ -3,44 +3,53 @@ name: Push to dev branch
...
-          node-version: "14"
-      - uses: ./.github/actions/functional-test
+          node-version: 16
+      - name: Install dependencies
+        run: npm ci
+      - name: Run tests
+        run: |
+          docker-compose up -d
+          # Start backend in background (& is mandatory to run process in background otherwise the task never continue)
+          npm run dev &
```
rolljee, line 50: Isn't the docker compose suppose to launch the whole task, why nom run dev is needed ?

.github/workflows/push_master.workflow.yml
```diff
@@ -11,36 +11,46 @@ jobs:
...
-          node-version: "14"
-      - uses: ./.github/actions/functional-test
+          node-version: 16
+      - name: Install dependencies
+        run: npm ci
+      - name: Run tests
+        run: |
+          docker-compose up -d
+          # Start backend in background (& is mandatory to run process in background otherwise the task never continue)
+          npm run dev &
```
rolljee, line 51: Same comment

docker-compose.yml
```diff
@@ -1,38 +1,81 @@
...
-    volumes:
-      - .:/var/app
     ports:
       - "7512:7512"
       - "9229:9229"
       - "1883:1883"
-    depends_on:
-      - redis
+    healthcheck:
+      test: ['CMD', 'curl', '-f', 'http://kuzzle:7512/_healthCheck']
```
rolljee, line 53: Since you are using Kuzzle 2.25, you can use "http://localhost:7512/_healthcheck" new public route

rolljee (changes requested):

.github/workflows/push_dev.workflow.yml
sebtiz13: Docker only run redis and elasticsearch, kuzzle is run out of docker to improve the log of errors in CI.
Also maybe it's better to run `npm run prod` instead development mode for better tests.

docker-compose.yml
sebtiz13: I think there is mistake because, I have used, this route for healthcheck on kuzzle service.

rolljee: http://kuzzle:7512/_healthCheck does exists indeed but it will be deprecated

.github/workflows/push_dev.workflow.yml
rolljee: This is not a good idea, kuzzle should be ran inside the docker compose, because when it will be deployed in production it will be used inside a docker container, so when you test it you should run it in the same environment. else your environment will not be the same while running inside your CI, witch is not what you want

If you seem to be lacking of logs, you should add a docker-compose logs kuzzle when the script fails in the wait-nuzzle-start.sh

sebtiz13: The problem is not especially for `wait-kuzzle.sh` but when functional tests cause errors.
IMHO it's better like that, but no problem, I can restore into docker.

docker-compose.yml
sebtiz13: Hum sorry, I Don't really understand the difference between "http://localhost:7512/_healthcheck" and "http://kuzzle:7512/_healthCheck" for the test.
The only difference is the DNS and I found it's more accurate to use the service name instead localhost which is less clear.

Could you explain to me the difference ?

rolljee: Sorry, I tested locally then forgot to update with the service name indeed

I meant http://kuzzle:7512/_healthcheck

The route you are using http://kuzzle:7512/_healthCheck is not a public route, it will be working as is until your kuzzle instance allow the anonymous the use this controller server.

```
{
--
verb: "get",
path: "/_healthCheck",
controller: "server",
action: "healthCheck",
}
```

Whereas the http://kuzzle:7512/_healthcheck and will always return a 200 since it is a public route

.github/workflows/push_dev.workflow.yml
rolljee: I understand why you want it this way, but in this way we diverge too much from the final environment in which a kuzzle is deployed and this creates more risk of trouble. Maybe we can talk about it on discord a bit to see if we can still have a docker environment, and more log that suit the debugging of the stack when it runs functional test

wdyt ?

docker-compose.yml
sebtiz13: Ok my bad, I just saw the difference in typo between `_healthCheck` and `_healthcheck`

sebtiz13: changed

docker-compose.yml
```diff
@@ -1,38 +1,81 @@
...
       - "9229:9229"
       - "1883:1883"
-    depends_on:
-      - redis
+    healthcheck:
+      test: ['CMD', 'curl', '-f', 'http://kuzzle:7512/_healthcheck']
+      timeout: 1s
+      interval: 2s
+      retries: 10
+      start_period: 30s
```
alexandrebouthinon, line 57: I think we should not set a Docker Compose healthcheck on Kuzzle since functional test for example could lead to permissions issues on this route

alexandrebouthinon: Or at least you should handle 403/401 status code to be allowed

sebtiz13: The healthcheck route are not forced to be public, or in minimal case, always allow call from localhost ?
Also, I don't see how to handle 403/401 status code here

alexandrebouthinon: The only way to do that is to create a custom healthcheck (like a Node.js or bash script for example)but for now ignore my comment

alexandrebouthinon (approved):

docker-compose.yml
rolljee: This health check method is public now 👌🏼 we should not have any issue see https://github.com/kuzzleio/kuzzle/pull/2455

PR #309, opened 2023-06-11
Fix composite measures should be correctly exported to CSV

### LINKED ISSUE #308: CSV export does not export all columns
When a measure is exported in CSV, only the property whose name is equal to the measure type is exported. Forgetting the potential others properties

## Expected Behavior
A measure with `position` with the properties `position` and `accuracy` should result in an export with 2 headers, one for each property.

## Current Behavior
At the moment, only the property `position` is present in the export.

tdislay:
## What does this PR do ?

Composite measures (measures with multiple properties, which can be objects) are now correctly exported. Before, only the property with the name equal to the measure type was exported (see the issue below).
Fixes #308.

### Other changes

* Property "position" was moved to the top of the `valueMappings` for the position measurement. It helps with consistency (which means, we do not need to sort properties from the mappings to display in a logical order the columns in the csv export).

### Boyscout

* Ordering some imports
* Corrected a misnamed test

### DIFF
--- a/lib/modules/measure/MeasureExporter.ts
+++ b/lib/modules/measure/MeasureExporter.ts
@@ -1,7 +1,7 @@
 import { randomUUID } from "node:crypto";
 import { PassThrough } from "node:stream";
 
-import _ from "lodash";
+import { stringify } from "csv-stringify/sync";
 import {
   InternalError,
   JSONObject,
@@ -9,12 +9,17 @@ import {
   NotFoundError,
   User,
 } from "kuzzle";
-import { stringify } from "csv-stringify/sync";
+import _ from "lodash";
 
-import { AskModelAssetGet, AskModelDeviceGet } from "../model";
+import {
+  AskModelAssetGet,
+  AskModelDeviceGet,
+  AskModelMeasureGet,
+} from "../model";
 import { DeviceManagerPlugin, InternalCollection } from "../plugin";
-import { ask, DigitalTwinContent } from "../shared";
+import { DigitalTwinContent, ask, flattenObject } from "../shared";
 
+import { NamedMeasures } from "../decoder";
 import { MeasureContent } from "./exports";
 
 type ExportOptions = {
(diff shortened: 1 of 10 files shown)


fmauNeko (approved):

thomas-mauran (approved):

lib/modules/model/types/ModelEvents.ts
```diff
@@ -15,3 +19,11 @@ export type AskModelDeviceGet = {
...
   result: DeviceModelContent;
 };
+
+export type AskModelMeasureGet = {
+  name: "ask:device-manager:model:measure:get";
+
+  payload: { type: string };
+
+  result: MeasureModelContent;
+};
```
sebtiz13, line 29: Nit-picking, why blank lines here ?

lib/modules/measure/MeasureExporter.ts
```diff
@@ -224,6 +229,66 @@ export class MeasureExporter {
...
+         */
+        const pathWithoutType = path.replace(new RegExp(type + "[.]?"), "");
+
+        /**
+         * @example
+         * 'temperature', 'acceleration.x', 'acceleration.accuracy'
+         */
+        const header = `${name}${
+          pathWithoutType !== "" ? `.${pathWithoutType}` : ""
+        }`;
```
sebtiz13, line 280:
```suggestion
        const header = `${name}.${pathWithoutType}`;
```

This ternary condition seems to be useless.
Unless if you want to handle nullish value in this case you have an error and the better solution is to use nullish coalsing, like `pathWithoutType ?? ''`

sebtiz13 (changes requested):

lib/modules/measure/MeasureExporter.ts
tdislay: You're not entirely right, it **was** useful, but your comment made me rework this crap. When I did it, I did not know how to do it in a more proper way.

I've fixed it. I know use a simple regex to do what I wanted to do.
I will add a comment in this PR explaining what I did

tdislay:
_Following my response to Seb's comment._

When I implemented this fix, I **assumed** that we should not repeat the type of a measure in the column header.
Normally the header of a measure should look like this (here I assume the measure's name !== measure's type in the decoder) : `temperatureInt.temperature, temperatureExt.temperature, accelerationSensor.acceleration.x, accelerationSensor.accuracy`.

What's actually going on is this: `temperatureInt, temperatureExt, accelerationSensor.x, accelerationSensor.accuracy`.

Do you guys, think this should not be this way ?

sebtiz13 (approved):

PR #311, opened 2023-07-17
Feat/assets groups last update

sebtiz13:
## What does this PR do ?

Improve assets groups to add the data `lastUpdate` to store when groups have been updated

  - Step 1 : update of a group from API
  - Step 2 : check `lastUpdate` property has been update


Aschen:
Why are you not using Kuzzle metadata? The `updatedAt` field contains the timestamp of the last modification of the document

PR #315, opened 2023-08-21
feat: digitaltwin exports

sebtiz13:
## What does this PR do ?

Add the list exports for digitalTwins

  - Step 1 : Call in `POST` the endpoint `/device-manager/<engineId>/<type>/_export`, with `query` and `sort` in body
  - Step 2 : Call the URL returned by previous request, you should get a CSV file

`<type>` is `assets` or `devices`

### Other changes

Refactor `MeasureExporter` with `AbstractExporter` to mutualize the exports logic

### DIFF
--- a/lib/modules/asset/AssetsController.ts
+++ b/lib/modules/asset/AssetsController.ts
@@ -2,6 +2,7 @@ import { ControllerDefinition, HttpStream, KuzzleRequest } from "kuzzle";
 
 import { MeasureExporter } from "../measure/";
 import { DeviceManagerPlugin, InternalCollection } from "../plugin";
+import { DigitalTwinExporter } from "../shared";
 
 import { AssetService } from "./AssetService";
 import { AssetSerializer } from "./model/AssetSerializer";
@@ -16,6 +17,7 @@ import {
 
 export class AssetsController {
   public definition: ControllerDefinition;
+  private exporter: DigitalTwinExporter;
   private measureExporter: MeasureExporter;
 
   constructor(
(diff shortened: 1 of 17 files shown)


lib/modules/asset/AssetsController.ts
```diff
@@ -230,17 +252,54 @@ export class AssetsController {
...
+        },
+      });
+
+      const stream = await this.exporter.sendExport(engineId, exportId);
+
+      return new HttpStream(stream);
+    }
+
+    const query = request.input.body?.query;
+    const sort = request.input.body?.sort;
```
tdislay, line 293:
```
const query = request.getBodyObject("query");
const sort = request.getBodyString("sort");
```

In order to correctly handle missing properties ?

lib/modules/device/DevicesController.ts
```diff
@@ -443,4 +463,40 @@ export class DevicesController {
...
+        },
+      });
+
+      const stream = await this.exporter.sendExport(engineId, exportId);
+
+      return new HttpStream(stream);
+    }
+
+    const query = request.input.body?.query;
+    const sort = request.input.body?.sort;
```
tdislay, line 489: same (getBodyString / Object)

lib/modules/device/DevicesController.ts
```diff
@@ -443,4 +463,40 @@ export class DevicesController {
...
     );
   }
+
+  async export(request: KuzzleRequest) {
+    const engineId = request.getString("engineId");
+
+    if (
+      request.context.connection.protocol === "http" &&
+      request.context.connection.misc.verb === "GET"
+    ) {
```
tdislay, line 473: Just asking, but why do we keep this controller with radically changing behavior (with no shared behavior) according to the type of connection ? Why don't we use two controllers ?

tdislay (approved):

lib/modules/asset/AssetsController.ts
sebtiz13: `query` and `sort` are objects and not mandatory.
This method allow optional properties while  `getBodyObject("query")` define it mandatory.
But it's true refactor it can be better, but need improvement in `kuzzle-sdk` on in `core`.

lib/modules/device/DevicesController.ts
sebtiz13: same

lib/modules/device/DevicesController.ts
sebtiz13: I'm not sure, but I think we need to improve core to can split this.
However, this behavior is due to the fact that it takes 2 type of request for an export.
First, we need to prepare export with `POST` request, then retrieve file with a `GET` request.

lib/modules/asset/AssetsController.ts
tdislay: my bad, misunderstood the code :+1:

PR #316, opened 2023-08-30
feat: improve export

sebtiz13:
## What does this PR do ?

- Add support for elasticsearch and koncorde language for export
- Improve handle error on export, return error message instead white page
- Limit export to 10k documents (prevent segfault with more than 25k results)

### DIFF
--- a/lib/modules/asset/AssetsController.ts
+++ b/lib/modules/asset/AssetsController.ts
@@ -200,12 +200,14 @@ export class AssetsController {
     const query = request.input.body?.query;
     const sort = request.input.body?.sort;
     const type = request.input.args.type;
+    const lang = request.getLangParam();
 
     const { measures, total } = await this.measureExporter.search(
       engineId,
       {
         endAt,
         id,
+        lang,
         query,
         sort,
         startAt,
@@ -251,6 +253,7 @@ export class AssetsController {
     const query = request.input.body?.query;
     const sort = request.input.body?.sort;
     const type = request.input.args.type;
+    const lang = request.getLangParam();
 
     const link = await this.measureExporter.prepareExport(
       engineId,
@@ -258,6 +261,7 @@ export class AssetsController {
       {
         endAt,
         id,
+        lang,
         query,
         sort,
         startAt,
@@ -291,11 +295,13 @@ export class AssetsController {
 
     const query = request.input.body?.query;
     const sort = request.input.body?.sort;
+    const lang = request.getLangParam();
 
     const link = await this.exporter.prepareExport(
       engineId,
       request.getUser(),
       {
+        lang,
         query,
         sort,
       }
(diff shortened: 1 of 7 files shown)


lib/modules/shared/services/DigitalTwinExporter.ts
```diff
@@ -25,33 +25,33 @@ export class DigitalTwinExporter extends AbstractExporter {
...
-      await this.sdk.ms.del(this.exportRedisKey(engineId, exportId));
-
-      return stream;
-    } catch (error) {
-      this.log.error(error);
-    }
+    const {
+      query,
+      sort,
+      lang = "elasticsearch",
```
rolljee, line 31: lang was set to koncorde by default before, we should keep the same otherwise it's a breaking change

lib/modules/asset/AssetsController.ts
```diff
@@ -227,20 +234,34 @@ export class AssetsController {
...
+        request.response.configure({
+          headers: {
+            "Content-Disposition": `attachment; filename="asset-${id}.csv"`,
+            "Content-Type": "text/csv",
+          },
+        });
+
+        return new HttpStream(stream);
+      } catch (error) {
+        request.response.configure({
```
rolljee, line 255: Why adding this instead of throwing a traditional error ?

rolljee (changes requested):

lib/modules/shared/services/AbstractExporter.ts
```diff
@@ -146,7 +147,7 @@ export abstract class AbstractExporter<P extends ExportParams = ExportParams> {
           stream.write(stringify([this.formatHit(columns, hit)]));
         }
 
-        result = await result.next();
+        result = result.fetched < 10000 ? await result.next() : null;
```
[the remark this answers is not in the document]
sebtiz13: Like test is after fetch, we does use `<` because `<=` cause one more turn of loop and finally fetch 10200 documents

lib/modules/shared/services/DigitalTwinExporter.ts
sebtiz13: It's true, but all other requests use `lang='elasticsearch'` by default, so it's inconsistent

lib/modules/asset/AssetsController.ts
sebtiz13: Because this endpoint is mostly called by browser and traditional error return a JSON that can be hard to understand for user.
Maybe the better solution could be to improve global error handler in the core

rolljee (approved):

lib/modules/asset/AssetsController.ts
tdislay: Add a comment to explain this in the code :)

lib/modules/shared/services/DigitalTwinExporter.ts
tdislay: Doesn't it already default to elasticsearch ? [See this](https://github.com/kuzzleio/kuzzle/blob/4e51e8cf5b3a0efcda0fc73416a5909cb4d01eb3/lib/api/request/kuzzleRequest.ts#L391C58-L391C58)

PR #317, opened 2023-09-28
feat/assets migrate tenant

thomas-mauran:
## assets/migrateTenant
The goal here is to provide a way to migrate assets and associated devices to a new tenant easily

  - Step 1 : clone and run
  - Step 2 : try the new route
  - Step 3: Check the new doc page

### Other changes
- Adding recoveryQueue from PaaS team to rollback the events
- Adding Ask functions to deviceService

### DIFF
--- a/lib/modules/asset/AssetService.ts
+++ b/lib/modules/asset/AssetService.ts
@@ -283,7 +283,7 @@ export class AssetService {
     return lock(`engine:${engineId}:${newEngineId}`, async () => {
       const recovery = new RecoveryQueue();
 
-      if (user.profileIds.includes("admin")) {
+      if (!user.profileIds.includes("admin")) {
         throw new BadRequestError(
           `User ${user._id} is not authorized to migrate assets`
         );
--- a/lib/modules/asset/AssetService.ts
+++ b/lib/modules/asset/AssetService.ts
@@ -460,6 +460,18 @@ export class AssetService {
             }
           }
         });
+
+        // clear the groups
+        await this.sdk.document.mUpdate<AssetContent>(
+          newEngineId,
+          InternalCollection.ASSETS,
+          assetsList.map((assetId) => ({
+            _id: assetId,
+            body: {
+              groups: [],
+            },
+          }))
+        );
       } catch (error) {
         await recovery.rollback();
         throw new BadRequestError(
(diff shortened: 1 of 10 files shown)


lib/modules/asset/AssetService.ts
```diff
@@ -268,6 +274,201 @@ export class AssetService {
...
+          .filter((asset) => asset._source.linkedDevices.length > 0)
+          .map((asset) => ({
+            assetId: asset._id,
+            linkedDevices: asset._source.linkedDevices,
+          }));
+
+        // Extra recovery step to relink back assets to their devices in case of rollback
+        recovery.addRecovery(async () => {
+          // Link the devices to the new assets
+          for (const asset of assetLinkedDevices) {
```
rolljee, line 355: Prefer loop for (let i = 0; i < ...; I++) as it is faster for a large amount of record

lib/modules/asset/AssetService.ts
```diff
@@ -268,6 +274,201 @@ export class AssetService {
...
+            const assetId = asset.assetId;
+            for (const device of asset.linkedDevices) {
+              const deviceId = device._id;
+              const measureNames = device.measureNames;
+              await ask<AskDeviceLinkAsset>(
```
rolljee, line 360: You should populate an array with promise and resolve them with promise.all, running await here will do them one by one. running promise will create a batch running them in // it will be much faster

lib/modules/asset/AssetService.ts
```diff
@@ -268,6 +274,201 @@ export class AssetService {
...
+                  measureNames: measureNames,
+                  user,
+                }
+              );
+            }
+          }
+        });
+
+        // detach from current tenant
+        for (const device of devices.hits) {
```
rolljee, line 375: Same comment about the for loop and the promise rather than await

lib/modules/asset/AssetService.ts
```diff
@@ -268,6 +274,201 @@ export class AssetService {
...
+        recovery.addRecovery(async () => {
+          await this.sdk.document.mCreate(
+            engineId,
+            InternalCollection.ASSETS,
+            assetsToMigrate
+          );
+        });
+
+        // Link the devices to the new assets
+        for (const asset of assetLinkedDevices) {
```
rolljee, line 431: Same comment here

rolljee (changes requested):
Good job overall, made some suggested changes. 

Also PR is missing tests :)

rolljee (approved):

PR #318, opened 2023-10-12
fix(asset_migration): fixes multiple assets migration and optimizations

Olive3034:
Fix: reuse mGet, mCreate, mDelete in order to maximize optimization process
Fix: Remove the recovery process that can also lead to unwanted things
Fix: Check if new asset already exists in new tenant instead of throwing an error
Fix: Check and allow to migrate an asset even if no linked device
Fix: returns an object containing 2 arrays: errors ids and successes ids (based on the mCreate result)
Optim: optimized calls (avoid update groups)
Optim: call sync/refresh at the end of the whole process
Optim: create all assets without linked devices in order to avoid inconsistent behavior. Only the newly created assets will then be used to link devices (the others are forgotten)


lib/modules/asset/AssetService.ts
```diff
@@ -304,213 +303,105 @@ export class AssetService {
...
-        await Promise.all(
-          devices.hits.map((device) => {
-            return ask<AskDeviceDetachEngine>(
+          // ... ant iterate over this list
+          for (const device of linkedDevices) {
+            // detach linked devices from current tenant (it also unkinks asset)
+            await ask<AskDeviceDetachEngine>(
               "ask:device-manager:device:detach-engine",
-              { deviceId: device._id, user }
+              { deviceId: device.id, user }
```
rolljee, line 348: can you keep this property to _id to be coherent with the document content

lib/modules/asset/AssetService.ts
```diff
@@ -304,213 +303,105 @@ export class AssetService {
...
       } catch (error) {
-        await recovery.rollback();
         throw new BadRequestError(
           `An error occured while migrating assets: ${error}`
         );
       }
+
+      if (!migrated) {
+        throw new BadRequestError(
+          `Error occured while migrating all the assets !`
```
rolljee, line 387: Instead of throwing this error, you can print all the assets that havent been migrated

lib/modules/asset/AssetService.ts
```diff
@@ -304,213 +303,105 @@ export class AssetService {
...
+        );
+      } else {
+        await this.sdk.collection.refresh(engineId, InternalCollection.ASSETS);
```
rolljee, line 390: You can make a for loop that iterate over each collection and refresh them, it will be cleaner

lib/modules/asset/AssetService.ts
rolljee: And maybe add the one the succeeded so we can print in the interface to the user the complete result of the process

rolljee: You can write a tests that fail to test that you print the right data

rolljee (changes requested):
2 points

- If you still want to remove the queue to rollback actions, you need to remove the associated file that came with it in the first place

- Also removing this queue rather to fixing it can lead to incoherent data inside ES in the long run

Olive3034:
> 2 points
> 
> * If you still want to remove the queue to rollback actions, you need to remove the associated file that came with it in the first place
> * Also removing this queue rather to fixing it can lead to incoherent data inside ES in the long run

* Sorry but what is "the associated file that came with it" ?

* In a first iteration, I prefer removing it rather than use something that can by itself add some more  inconsistencies. If the normal process has failed, why the recovery queue operation can't also fail, and add more inconsistencies ?
It is also a choice to iterate one asset by one asset, to avoid a large amount of inconsitencies (vs mCreate / mDelete).

lib/modules/asset/AssetService.ts
Olive3034: In that precise cas, I want to throw an error when ALL assets have failed to migrate.
An improvement should be to return errors [] / successes [] to know which ones have been successfully migrated, and the ones that failed.
For the test, yes ... but it will spend lots of hours to me ;-)

lib/modules/asset/AssetService.ts
Olive3034: yes, I ll modify it.

lib/modules/asset/AssetService.ts
rolljee: You can do something like 

```js
const collectionsToRefresh = [
          {
            collection: InternalCollection.ASSETS,
            engine: engineId,
          },
          {
            collection: InternalCollection.DEVICES,
            engine: engineId,
          },
          {
            collection: InternalCollection.ASSETS,
            engine: newEngineId,
          },
          {
            collection: InternalCollection.DEVICES,
            engine: newEngineId,
          },
          {
            collection: InternalCollection.DEVICES,
            engine: this.config.adminIndex,
          },
        ].map(({ engine, collection }) => {
          return this.sdk.collection.refresh(engine, collection);
        });

        await Promise.all(collectionsToRefresh);
   ```
   
   The difference between the screenshot and your code is that running Promise.all will launch all request in parallel.
   
   Your code process refresh one by one, so the code runs faster using promise.all because you don't wait for the process to end before launching a new one.
   
   In javascript this is the main difference will using await vs promises.
   
   Here you don't need to await each refresh individually you can trigger them all at once.

File: lib/modules/asset/AssetService.ts (hunk 2 above)
Comment (rolljee) [reply lib/modules/asset/AssetService.ts:387]: If you don't want to take the time now to have such arrays to print errors and success migration, you should add an issue in this repository that talk about this improvements that should be done.

Also we should not discuss about how much time it will take to add a test, because if at some point we have to change this function behaviour and we made a regression about this part of the code. It will be much harder to debug in 6 months because we will have forgotten what we written here. leading to much more hours in debugging that the time spent to write a single test

**Review by rolljee — CHANGES_REQUESTED** [review]

**Comment by rolljee** [comment]
> * Sorry but what is "the associated file that came with it" ?

The associated file is RecoveryQueue in `lib/modules/shared/utils/recoveryQueue.ts`

> * In a first iteration, I prefer removing it rather than use something that can by itself add some more  inconsistencies. If the normal process has failed, why the recovery queue operation can't also fail, and add more inconsistencies ?

That what I was saying, by taking the time to fix this behaviour of the recoveryQueue. But it's true that there is no such process in the whole plugin so we might wait another iteration to implement it correctly

>   It is also a choice to iterate one asset by one asset, to avoid a large amount of inconsitencies (vs mCreate / mDelete).

The process will be much longer but less errors prone, if we want KDM to be more performant we should still use mCreate and mDelete requests and running migration in batch. We should raise an issue about this performance cost choice

**Review by fmauNeko — CHANGES_REQUESTED** [review]
Switching from 1 `mGet` call to Iterating over N assets using `exists` (which under the hood is a `get`) is a huge deoptimization and will probably sink the performance.
Same goes for 1 `mCreate` to N `create`.
I think this will require a full test and benchmark suite before we consider merging it.

File: lib/modules/asset/AssetService.ts (hunk 3 above)
Comment (Olive3034) [reply lib/modules/asset/AssetService.ts:390]: Understood. Great !

**Comment by Olive3034** [comment]
> Switching from 1 `mGet` call to Iterating over N assets using `exists` (which under the hood is a `get`) is a huge deoptimization and will probably sink the performance. Same goes for 1 `mCreate` to N `create`. I think this will require a full test and benchmark suite before we consider merging it.

Ye, I'll do some benchmarks and give results to see if acceptable.
But, I prefer a slower solution 100% successfully.
Currently, when we have a fail, we lead into a lots of inconsistencies.
And the first point is to fix this.

File: lib/modules/asset/AssetService.ts (hunk 2 above)
Comment (Olive3034) [reply lib/modules/asset/AssetService.ts:387]: I'll add an issue for this, in order to implement it later, in a next iteration

**Comment by Olive3034** [comment]
I don't understand a TF (export csv) is not passing anymore ...

**Comment by Olive3034** [comment]
> Switching from 1 `mGet` call to Iterating over N assets using `exists` (which under the hood is a `get`) is a huge deoptimization and will probably sink the performance. Same goes for 1 `mCreate` to N `create`. I think this will require a full test and benchmark suite before we consider merging it.

This is a new feature, and I prefer a slower solution 100% successfully.
Currently, when we have a fail, we lead into a lots of inconsistencies, and this is the first point to fix.
@fmauNeko Also, I have removed the iteration with exists, and preffered mGet calls.

File: lib/modules/asset/AssetService.ts [hunk 4]
```diff
@@ -290,7 +289,7 @@ export class AssetService {
       }
 
       try {
-        // check if tenant destination of the the same group
+        // check if tenant destination is of the same group
```
rolljee, line 292:
```suggestion
        // check if tenant destination is in the same group
```

lib/modules/asset/AssetService.ts
```diff
@@ -304,213 +303,121 @@ export class AssetService {
...
         );
-
-        if (existingAssets.successes.length > 0) {
-          throw new BadRequestError(
-            `Assets ${existingAssets.successes
-              .map((asset) => asset._id)
-              .join(", ")} already exists in engine ${newEngineId}`
-          );
+        const assetsIdInDest = assetsInDest.successes.map((a) => a._id);
+        //And if yes, remove them from the list to use
```
rolljee, line 313: This comment can be removed

rolljee (approved):
While this implementation should be reworked to be faster in the futur, we could implement it as is for now

Shiranuit (changes requested):
IMHO, this looks like a huge deoptimisation instead of an optimisation like it's stated in the title.

replacing the mCreate with N calls to create is going to cause a huge degredation of performances. this is likely to impact every users of KDM, even more so if the document are relatively large.
The same idea is applied for every operation done in the loops.

I understand that you're trying to have something that produces more consistent results but maybe fixing the RecoveryQueue might have been a been a better idea before introducing such regressions.

At least what could have been done to reduced the number of incosistent result and kept a decent speed by using mCreate, mDelete etc on smaller batches, for example if you had 100 documents you could have splitted them in batches of 10 documents, so you would have 10 calls to mCreate that each contains 10 documents, and you would stop if one of the document in the batch had an error, this way you would have kept a somewhat decent speed (the 10 document size is an example) while avoiding having a huge load of document in an inconsistent state.

It important to be able to measure the impact of such changes and so having benchmarks in those cases are mandatory since this will pretty much affect any production using KDM.

Olive3034:
> While this implementation should be reworked to be faster in the futur, we could implement it as is for now

I think I have rewritten everything to use mGet, mCreate and mDelete, but taking care to avoid several inconsistencies

Olive3034:
> IMHO, this looks like a huge deoptimisation instead of an optimisation like it's stated in the title.
> 
> replacing the mCreate with N calls to create is going to cause a huge degredation of performances. this is likely to impact every users of KDM, even more so if the document are relatively large. The same idea is applied for every operation done in the loops.
> 
> I understand that you're trying to have something that produces more consistent results but maybe fixing the RecoveryQueue might have been a been a better idea before introducing such regressions.
> 
> At least what could have been done to reduce the number of incosistent result and keep a decent speed by using mCreate, mDelete etc on smaller batches, for example if you had 100 documents you could have splitted them in batches of 10 documents, so you would have 10 calls to mCreate that each contains 10 documents, and you would stop if one of the document in the batch had an error, this way you would have kept a somewhat decent speed (the 10 document size is an example) while avoiding having a huge load of document in an inconsistent state.
> 
> It important to be able to measure the impact of such changes and so having benchmarks in those cases are mandatory since this will pretty much affect any production using KDM.

I have rewriten keeping back the use of all mGet, mCreate and mDelete to improve performances, butbut while minimizing the risk of inconsistencies.

rolljee (approved):
This is better 👍🏼 

While we should also make attach/detach/link/unlink device working in batch to make this a proper efficient migration process, this could be done in a futur step.

Could you raise an issue in this repository for that as well ?

I believe this could be merge

lib/modules/asset/AssetService.ts
```diff
@@ -279,239 +279,157 @@ export class AssetService {
...
-        if (assetsList.length === 0) {
-          throw new BadRequestError("No assets to migrate");
-        }
+      //Get all assets to migrate
+      const assets = await this.sdk.document.mGet<AssetContent>(
+        engineId,
+        InternalCollection.ASSETS,
+        assetsList
+      );
+      errors.push(...assets.errors);
```
Shiranuit, line 314: FYI, doing a push with a spread operator `...` on an array is really bad, the reason for this is when you're using the spread operator like in the following example

```js
x=[];
y=[1,2,3,4];

x.push(...y);
```

It's the equivalent on creating a `push` function that takes 4 parameters, which could be represented like so

```js
x=[]
y=[1,2,3,4]

x.push(y[0], y[1], y[2], y[3]);
```

But here 4 parameters is not going to do any harm but lets say there is 50 elements in the array, now Javascript as to create a new function push at runtime that has to take 50 parameters

```js
x=[]
y=[1,2,3,4,5,...,49,50]

x.push(y[0], y[1], y[2], y[3],...,y[48],y[49]);
```

Now imagine it has to do this each time an array with a different size is used.

This is a not really what is done because V8 does really complex stuff to optimise this, but this give a good perspective to what might happen.

Instead of doing this you should use the `concat` function that takes an array and merge both array into a new one.


Here is a microbenchmark to show you the impact

```
Array concat function x 1,989,683 ops/sec ±4.84% (79 runs sampled)
Spread operator x 195,559 ops/sec ±1.22% (92 runs sampled)
Iterate over array and push element one by one x 241,256 ops/sec ±0.55% (95 runs sampled)
Fastest is Concat
```

This applies for every  instance where you used the spread operator with what might contain lots of data

lib/modules/asset/AssetService.ts
```diff
@@ -279,239 +279,157 @@ export class AssetService {
...
+      for (const asset of assetsCreated.successes) {
+        //We need to recover the linked devices to this asset,
+        //because we reset them when we create the asset
+        const assetOriginal = assets.successes.find((a) => a._id === asset._id);
+        if (!assetOriginal) {
+          this.context.log.error(
+            `Impossible to recover the original asset with id = ${asset._id}`
           );
+          continue;
         }
```
Shiranuit, line 360: This condition is never reachable since `assets.successes` contains all the assets that you've get before, and `assetsCreated` is the result of the call to `mCreate` based on the data obtained with `assets.successes` meaning that if an assets as successfuly been created it will always be contained  in the `assets.successes` thus assetOriginal is always going to exists and this condition is not useful.

Shiranuit (changes requested):

lib/modules/asset/AssetService.ts
Olive3034: Yes, indeed. I know... but as it can programmatically reach undefined, I have keep this test it anyway.
I can remove it.

lib/modules/asset/AssetService.ts
Olive3034: ok

lib/modules/asset/AssetService.ts
```diff
@@ -279,239 +279,157 @@ export class AssetService {
...
-          assetsList
-        );
+      //Get all the asset content, in order to create thm by batch
+      const assetsContent = assets.successes.map((asset) => ({
+        _id: asset._id,
+        body: asset._source,
+      }));
+
+      //We want to create the new asset with linked devices and groups empty
+      const assetsContentCopy = _.cloneDeep(assetsContent);
```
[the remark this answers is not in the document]
Olive3034: OKi, but what is the issue using LoDash ? It is also used at several places in the project ...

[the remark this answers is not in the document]
rolljee: There is not much, it is just that when there is a native implementation of a function, we should use it rather than using a version from a lib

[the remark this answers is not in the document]
Olive3034: unfortunately, structuredClone makes the TF to fail :( 
PluginImplementationError: Caught an unexpected plugin error: structuredClone is not defined
So maybe still using lodash deepClone ?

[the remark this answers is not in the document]
fmauNeko: `deepClone` actually has better performance for primitive objects than `structuredClone`
image

PR #322, opened 2023-10-30
feat(digitalTwin): allow to modify mapping from application

sebtiz13:
## What does this PR do ?

Allow to modify digital twin mapping from application.
These allow to extend mapping properties from application to use custom properties in documents.

This is needed to add `softTenant` property to restrict access to a specific soft tenant, this mapping modification is added from package iot-platform-backend package with plugins setup.

### DIFF
--- a/tests/application/app.ts
+++ b/tests/application/app.ts
@@ -15,6 +15,17 @@ const app = new Backend("kuzzle");
 
 const deviceManager = new DeviceManagerPlugin();
 
+//? Add custom mapping properties
+deviceManager.config.engineCollections.asset.mappings.properties["softTenant"] = {
+  type: "keyword",
+  fields: { text: { type: "text" } },
+};
+deviceManager.config.engineCollections.device.mappings.properties["softTenant"] = {
+  type: "keyword",
+  fields: { text: { type: "text" } },
+};
+
+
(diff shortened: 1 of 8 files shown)


OlivierCavadenti (approved):

tests/application/app.ts
```diff
@@ -15,6 +15,17 @@ const app = new Backend("kuzzle");
 
 const deviceManager = new DeviceManagerPlugin();
 
+//? Add custom mapping properties
+deviceManager.config.engineCollections.asset.mappings.properties["softTenant"] = {
+  type: "keyword",
+  fields: { text: { type: "text" } },
```
etrousset, line 21: Why do we need to index this fiald as "text" ? Keyword should be enough, no?

tests/application/app.ts
```diff
@@ -15,6 +15,17 @@ const app = new Backend("kuzzle");
...
+};
+deviceManager.config.engineCollections.device.mappings.properties["softTenant"] = {
+  type: "keyword",
+  fields: { text: { type: "text" } },
```
etrousset, line 25: Why do we need to index this fiald as "text" ? Keyword should be enough, no?

etrousset (changes requested):

PR #323, opened 2023-12-29
Kzlprd 166 upsert assets

Leagian:
## What does this PR do ?

This PR adds an upsert method in the AssetService. The upsert method is designed to update an existing asset with new metadata, or create a new asset if one does not already exist with the specified assetId.

 npm run test:functional

### DIFF
--- a/lib/modules/asset/AssetService.ts
+++ b/lib/modules/asset/AssetService.ts
@@ -138,57 +138,33 @@ export class AssetService extends BaseService {
   /**
    * Update or Create an asset metadata
    */
-  /**
-   * Update or Create an asset metadata
-   */
   public async upsert(
     engineId: string,
     assetId: string,
+    model: string,
+    reference: string,
     metadata: Metadata,
     request: KuzzleRequest
   ): Promise<KDocument<AssetContent>> {
     return lock(`asset:${engineId}:${assetId}`, async () => {
-      let asset = null;
-      try {
-        asset = await this.get(engineId, assetId, request);
-      } catch (error) {
-        this.app.log.error(`Asset ${assetId} not found`);
-        asset = {
-          _id: assetId,
-          _source: {
-            groups: [],
-            lastMeasuredAt: 0,
-            linkedDevices: [],
-            measures: undefined,
-            metadata: {},
-            model: "",
-            reference: "",
-          },
-        };
+      const asset = await this.get(engineId, assetId, request).catch(
+        () => null
+      );
+
+      if (!asset) {
+        return this.create(engineId, model, reference, metadata, request);
       }
 
-      const upsertedPayload = await this.app.trigger<EventAssetUpdateBefore>(
+      const updatedPayload = await this.app.trigger<EventAssetUpdateBefore>(
         "device-manager:asset:update:before",
         { asset, metadata }
       );
 
-      const _source = asset
-        ? { ...asset._source, metadata: upsertedPayload.metadata }
-        : {
-            groups: [],
-            lastMeasuredAt: 0,
-            linkedDevices: [],
-            measures: undefined,
-            metadata: upsertedPayload.metadata,
-            model: "",
-            reference: "",
-          };
-
-      const upsertedAsset = await this.upsertDocument<AssetContent>(
+      const updatedAsset = await this.updateDocument<AssetContent>(
         request,
         {
           _id: assetId,
-          _source,
+          _source: { metadata: updatedPayload.metadata },
         },
         {
           collection: InternalCollection.ASSETS,
@@ -199,14 +175,14 @@ export class AssetService extends BaseService {
 
       await this.assetHistoryService.add<AssetHistoryEventMetadata>(engineId, [
         {
-          asset: upsertedAsset._source,
+          asset: updatedAsset._source,
           event: {
             metadata: {
-              names: Object.keys(flattenObject(upsertedPayload.metadata)),
+              names: Object.keys(flattenObject(updatedPayload.metadata)),
             },
             name: "metadata",
           },
-          id: upsertedAsset._id,
+          id: updatedAsset._id,
           timestamp: Date.now(),
         },
       ]);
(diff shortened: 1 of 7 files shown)


lib/modules/asset/AssetService.ts
```diff
@@ -135,6 +135,97 @@ export class AssetService extends BaseService {
     });
   }
 
+  /**
+   * Update or Create an asset metadata
+   */
+  /**
+   * Update or Create an asset metadata
```
fmauNeko, line 142: Duplicate comment

lib/modules/asset/AssetService.ts
```diff
@@ -135,6 +135,97 @@ export class AssetService extends BaseService {
...
+    assetId: string,
+    metadata: Metadata,
+    request: KuzzleRequest
+  ): Promise<KDocument<AssetContent>> {
+    return lock(`asset:${engineId}:${assetId}`, async () => {
+      let asset = null;
+      try {
+        asset = await this.get(engineId, assetId, request);
+      } catch (error) {
+        this.app.log.error(`Asset ${assetId} not found`);
```
fmauNeko, line 155: In that case this is not really an error, it should be an info or debug message

lib/modules/asset/AssetService.ts
```diff
@@ -135,6 +135,97 @@ export class AssetService extends BaseService {
...
+          },
+        };
+      }
+
+      const upsertedPayload = await this.app.trigger<EventAssetUpdateBefore>(
+        "device-manager:asset:update:before",
+        { asset, metadata }
+      );
+
+      const _source = asset
```
fmauNeko, line 175: That ternary is unnecessary because asset is defined in the catch L157, so its condition will always be truthy.
`const _source = { ...asset._source, metadata: upsertedPayload.metadata }` should be enough in that case.

fmauNeko (changes requested):
Could you add some tests too, please ?

lib/modules/asset/AssetService.ts
Juiced66: It wasn't because asset can be null in case it do not exists

lib/modules/asset/AssetService.ts
```diff
@@ -135,6 +135,97 @@ export class AssetService extends BaseService {
...
+            measures: undefined,
+            metadata: {},
+            model: "",
+            reference: "",
+          },
+        };
+      }
+
+      const upsertedPayload = await this.app.trigger<EventAssetUpdateBefore>(
+        "device-manager:asset:update:before",
```
Juiced66, line 171: Should be "device-manager:asset:upsert:before" .  Not sure there is corresponding event type, in this case we should prob add it.

lib/modules/asset/AssetService.ts
```diff
@@ -135,6 +135,97 @@ export class AssetService extends BaseService {
...
+            },
+            name: "metadata",
+          },
+          id: upsertedAsset._id,
+          timestamp: Date.now(),
+        },
+      ]);
+
+      await this.app.trigger<EventAssetUpdateAfter>(
+        "device-manager:asset:update:after",
```
Juiced66, line 215: Same as Line 171

lib/modules/shared/services/BaseService.ts
```diff
@@ -154,6 +155,50 @@ export abstract class BaseService {
...
+  ): Promise<KDocument<T>> {
+    const kuzzleRequest = this.normalizeKuzzleRequest(request, {
+      collection,
+      engineId,
+    });
+    const user = kuzzleRequest.getUser();
+    const refresh = kuzzleRequest.getRefresh();
+    const retryOnConflict = options.retryOnConflict || false;
+
+    const [modifiedDocument] = await this.app.trigger<
```
Juiced66, line 181: @fmauNeko should it also trigger update pipes ?

lib/modules/asset/AssetService.ts
rolljee: This is marked as resolved but this is still update event, it should be upsert

lib/modules/asset/AssetService.ts
rolljee: This is marked as resolved but this is still update event, it should be upsert

lib/modules/asset/AssetsController.ts
```diff
@@ -37,6 +38,12 @@ export class AssetsController {
           handler: this.create.bind(this),
           http: [{ path: "device-manager/:engineId/assets", verb: "post" }],
         },
+        upsert: {
+          handler: this.upsert.bind(this),
+          http: [
+            { path: "device-manager/:engineId/assets/:_id", verb: "post" },
```
rolljee, line 44: This path does not reflect the behaviour of this action

rolljee (changes requested):

lib/modules/asset/AssetService.ts
Leagian: After the comments, I then attempted to use the upsertDocument method in the BaseService file, but it wasn't appropriate because in the Asset service file, if you use the upsertDocument method, you would need to reconstruct the logic of the existing create and update methods. Therefore, with the aim of maintaining cleanliness and logic in the upsert method within the AssetService file, I chose to use the already established create and update methods

lib/modules/asset/AssetsController.ts
Leagian: I used POST by following this : https://docs.kuzzle.io/core/2/api/controllers/document/upsert/
Should I change it to PUT?

rolljee: Nah, you are right I though the path was the same as the update action, but nvm it's okay :)

rolljee (approved):

lib/modules/asset/AssetsController.ts
Leagian: no problem !

fmauNeko (approved):

PR #328, opened 2024-01-26
KZLPRD-233 feat: 🎸 added action name on list decoders API response

etrousset:
✅ Closes: KZLPRD-233

As the action for a decoder can be different from the deviceModel, added the proper action to call for a deviceModel in the list response to ease integration in third party tools

image


rolljee (changes requested):
This need to be added in the related test 👍🏼

rolljee (approved):

PR #334, opened 2024-03-14
Fix(migrate_tenant_sanity_check) : raise an error

Olive3034:
Raise an error if all assets to migrate tenants already exists in tenant destination


Olive3034:
> Missing tests
@rolljee , should be done now ;-)

rolljee (approved):
Purrrrfect !

PR #342, opened 2024-05-17
feat: add UTC export by default for devices and assets

thomas-mauran:
## What does this PR do ?
WIP

  - Step 1 :
  - Step 2 :
  - Step 3 :

### Other changes

KZLPRD-330


afondard (approved):

Juiced66:
Should this take a timezone also? 
Do we need to keep both timestamp and formatted time ?

etrousset:
> Should this take a timezone also? Do we need to keep both timestamp and formatted time ?
As we are exporting raw UTC time, the timezone is out of scop of this PR
Yes, we need to keep raw timestamp on one hand, the formatted time is for convenience so that it is easier for human to analyze the data

etrousset (approved):

PR #343, opened 2024-05-22
feat(measuremodels): add optional local names and unit to measure definitions

QuentinRousselet:
## What does this PR do ?

This PR to give the possibility to add friendly names and units internationalized in the measure model definition.
It also adds friendly names and units to default measure models

1- On the IOT platform, register a measure via the instance of the app like so :

``` typescript
export const measureDefinition: MeasureDefinition = {
  valuesMappings: { temperature: { type: "float" } },
  valuesDetails: {
    temperature: {
      en: {
        friendlyName: "Temperature",
        unit: "°C",
      },
      fr: {
        friendlyName: "Température",
        unit: "°C",
      },
    },
  },
};


app.plugin.get<DeviceManagerPlugin>('device-manager');

  deviceManager.models.registerMeasure('measureName', measureDefinition);
```
2- On the admin console, in the platform index, models collection, verify that the document corresponding to your measureName has been created with the correct valuesDetails fields.
image

### DIFF
--- a/lib/modules/measure/types/MeasureDefinition.ts
+++ b/lib/modules/measure/types/MeasureDefinition.ts
@@ -5,23 +5,24 @@ import { JSONObject } from "kuzzle-sdk";
  *
  * @example
  * {
- *    locales: {
- *              en:{
- *                  name:"Temperature",
- *                  unit:"°C"
- *              },fr:{
- *                  name:"Température",
- *                  unit:"°C"
- *              }
- *    },
- * }
+ *  en: {
+ *    friendlyName: "Temperature",
+ *    unit: "°C",
+ *  },
+ *  fr: {
+ *    friendlyName: "Température",
+ *    unit: "°C",
+ *  },
+ *}
  */
+
 interface MeasureLocales {
   [localeString: string]: {
     friendlyName: string;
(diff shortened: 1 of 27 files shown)


lib/modules/measure/types/MeasureDefinition.ts
```diff
@@ -1,16 +1,55 @@
...
  *
  * @example
  * {
  *   valuesMappings: { temperature: { type: 'float' } },
+ *   valuesDetails: {
+ *        temperature:{
+ *              en:{
+ *                  name:"Temperature",
+ *                  unit:"°C"
+ *              },fr:{
```
Juiced66, line 39: lint

lib/modules/measure/types/MeasureDefinition.ts
```diff
@@ -1,16 +1,55 @@
 import { JSONObject } from "kuzzle-sdk";
 
+/* *
+ * Represents a measure information and localization
+ *
+ * @example
+ * {
+ *    locales: {
```
Juiced66, line 8: looks like there's no `locales` property ?

alexandrebouthinon (approved):

PR #345, opened 2024-06-03
fix(models):  list asset returns commons assets

Juiced66:
## What does this PR do ?

- [Exposes](https://github.com/kuzzleio/kuzzle-device-manager/actions/runs/9345460608) and fix a bug were listAssets would not return commons templates on a given engine

- create a engine : commons model
- try to list on an engine
- it should now return commons templates

### Other changes
- added a test

### DIFF
--- a/tests/scenario/migrated/model-controller.test.ts
+++ b/tests/scenario/migrated/model-controller.test.ts
@@ -127,6 +127,23 @@ describe("features/Model/Controller", () => {
     await expect(promise).rejects.toMatchObject({ status: 404 });
   });
 
+  it("List on an engine is also returning commons assets models", async () => {
+    const response = await sdk.query({
+      controller: "device-manager/models",
+      action: "listAssets",
+      engineGroup: "engine-ayse",
+    });
+
+    expect(response.result).toMatchObject({
+      total: 3,
+      models: [
+        { _id: "model-asset-Container" },
+        { _id: "model-asset-Plane" },
+        { _id: "model-asset-Warehouse" },
+      ],
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


afondard (approved):

QuentinRousselet (approved):

tests/scenario/migrated/model-controller.test.ts
```diff
@@ -127,6 +127,23 @@ describe("features/Model/Controller", () => {
...
+
+    expect(response.result).toMatchObject({
+      total: 3,
+      models: [
+        { _id: "model-asset-Container" },
+        { _id: "model-asset-Plane" },
+        { _id: "model-asset-Warehouse" },
+      ],
+    });
+  });
```
sebtiz13, line 145: IMHO this test could be improved
Your test case check the same thing of the previous test (`Write and List an Asset model`).
```js
    response = await sdk.query({
      controller: "device-manager/models",
      action: "listAssets",
      engineGroup: "commons",
    });
```

So that doesn't really test if commons templates are included with engine models or if it's return instead of.
Could you add a specific engine models to ensure, that is correctly returned with commons models ?

Juiced66: All models are commons models but yeah not perfect maybe I should declare a new model on an engine to. 

List is used with common in the other test and with an engine there

Juiced66: done, use other-group in place of ayse that is in commons group

rolljee (approved):

sebtiz13 (approved):

PR #347, opened 2024-06-07
build(types): use semantic-release for types package

sebtiz13:
## What does this PR do ?

Create `semantic-release` plugin to release the type package instead of having to do it manually.
Also improve type package release delivery by handle beta release tag (not handled actually), that avoid to use `latest` tag when release a beta.


etrousset (approved):

rolljee (approved):

alexandrebouthinon (approved):
Noice! 👌

PR #355, opened 2024-07-17
feat: upsert devices

Juiced66:
Add an upsert method to devices api

### DIFF
--- a/lib/modules/device/DevicesController.ts
+++ b/lib/modules/device/DevicesController.ts
@@ -26,6 +26,7 @@ import {
   ApiDeviceUnlinkAssetResult,
   ApiDeviceUpdateResult,
   ApiDeviceGetMeasuresResult,
+  ApiDeviceUpsertResult,
 } from "./types/DeviceApi";
 
 export class DevicesController {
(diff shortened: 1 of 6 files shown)


lib/modules/device/DevicesController.ts
```diff
@@ -56,6 +57,9 @@ export class DevicesController {
             { path: "device-manager/:engineId/devices/:_id", verb: "put" },
           ],
         },
+        upsert: {
+          handler: this.upsert.bind(this),
+        },
```
afondard, line 62: It looks like it is missing the http configuration

afondard (changes requested):
Adding the documentation for the upsert may be a good idea?

lib/modules/device/DevicesController.ts
Juiced66: oupsyy ty

Juiced66: done

fmauNeko (approved):

afondard (approved):

etrousset (approved):

PR #358, opened 2024-08-06
feat(metadata): add option field readOnly to prevent editing

cyrng:

  - Step 1 :
  - Step 2 :
  - Step 3 :

### DIFF
--- a/lib/modules/model/types/ModelContent.ts
+++ b/lib/modules/model/types/ModelContent.ts
@@ -38,6 +38,7 @@ export interface MetadataDetails {
     locales: {
       [locale: string]: LocaleDetails;
     };
+    readOnly?: boolean;
   };
 }
 


lib/modules/model/types/ModelContent.ts
```diff
@@ -38,6 +38,7 @@ export interface MetadataDetails {
     locales: {
       [locale: string]: LocaleDetails;
     };
+    readOnly?: boolean;
```
Kuruyia, line 41: Instead of having this field optional, wouldn't it be more appropriate to have a sensible default value for this?

cyrng: Yes I have set the default value to false in the frontend if the field does not exist. If the field is not optional, there will be the need to update the metadataDetails mapping for every existing asset/device.

Kuruyia (approved):

tests/scenario/migrated/model-controller.test.ts
sebtiz13, line 1: Maybe It would be good to test that the field is correctly retrieved after creating an asset and a device with this parameter in the model

afondard (approved):

PR #361, opened 2024-08-12
ci: add rebase on end of release process

sebtiz13:
## What does this PR do ?

Simply add rebase on end of release process to keep linear history

### DIFF
--- a/.github/workflows/release.workflow.yaml
+++ b/.github/workflows/release.workflow.yaml
@@ -1,4 +1,4 @@
-name: Push checks
+name: Release
 
 on:
   push:
@@ -16,10 +16,10 @@ jobs:
       pull-requests: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Setup Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: "lts/*"
           registry-url: "https://registry.npmjs.org"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.workflow.yaml
```diff
@@ -38,6 +38,12 @@ jobs:
           SEMANTIC_RELEASE_SLACK_WEBHOOK: ${{ secrets.SEMANTIC_RELEASE_SLACK_WEBHOOK }}
         run: npx semantic-release
 
+      - name: Rebase 2-dev (keep linear history)
+        run: |
+          git fetch --no-tags origin 2-dev:2-dev
+          git rebase ${GITHUB_REF_NAME} 2-dev
+          git push origin 2-dev
```
alexandrebouthinon, line 45: Will not work since it require a force push (commit IDs are refresh each time you rebase). Note that the CI user will become the author/co-author of each rebased commit

etrousset (changes requested):
According to Aex B comment, this needs fixing?

.github/workflows/release.workflow.yaml
sebtiz13: Here that just, add the commit `chore(release): x.x.x-beta.y [skip ci]` created by **semantic-release** to the dev branch, as what is done in other repositories

Juiced66: there's no force push as it is fast forward in principle

Juiced66 (approved):

PR #367, opened 2024-08-30
fix: only list asset models for the requested engine group

Kuruyia:
## What does this PR do ?

This fixes a regression introduced in #362 on the `listAssets` (and by extension, `searchAssets`) actions of the `device-manager/models` controller that wasn't correctly taking into account the requested engine group and was returning the models for every group instead.

The problematic Elasticsearch query has been fixed.

## Other changes

Two new asset models, `Room` and `StreetLamp`, were added in the test application, respectively to the `air_quality` and `public_lighting` engine groups.

Tests were added to avoid this regression in the future.

## Boyscout

To avoid an inconsistent list of asset models, the asset models are now reset before each test of the `asset-model.test.ts` file.

### DIFF
--- a/tests/application/assets/Room.ts
+++ b/tests/application/assets/Room.ts
@@ -1,0 +1,18 @@
+import { AssetModelDefinition } from "../../../index";
+
+export const roomAssetDefinition: AssetModelDefinition = {
+  measures: [
+    {
+      name: "temperature",
+      type: "temperature",
+    },
+    {
+      name: "humidity",
+      type: "humidity",
+    },
+  ],
+  metadataMappings: {
+    floor: { type: "integer" },
+    width: { type: "integer" },
+  },
+};
--- a/tests/application/assets/Room.ts
+++ b/tests/application/assets/Room.ts
@@ -10,6 +10,14 @@ export const roomAssetDefinition: AssetModelDefinition = {
       name: "humidity",
       type: "humidity",
     },
+    {
+      name: "co2",
+      type: "co2",
+    },
+    {
+      name: "illuminance",
+      type: "illuminance",
+    },
   ],
   metadataMappings: {
     floor: { type: "integer" },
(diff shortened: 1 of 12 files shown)


Juiced66 (approved):

QuentinRousselet (approved):

tests/application/assets/Room.ts
```diff
@@ -0,0 +1,18 @@
...
+  measures: [
+    {
+      name: "temperature",
+      type: "temperature",
+    },
+    {
+      name: "humidity",
+      type: "humidity",
+    },
+  ],
```
afondard, line 13: You could define the same measures as in the Room asset model from [iot-platform](https://github.com/kuzzleio/iot-platform/blob/3-dev/apps/api/lib/modules/air_quality/assets/Room.ts)
```suggestion
  measures: [
    {
      name: "temperature",
      type: "temperature",
    },
    {
      name: "humidity",
      type: "humidity",
    },
    {
      name: "co2",
      type: "co2",
    },
    {
      name: "illuminance",
      type: "illuminance",
    },
  ],
```

tests/application/assets/StreetLamp.ts
```diff
@@ -0,0 +1,13 @@
...
+export const streetLampAssetDefinition: AssetModelDefinition = {
+  measures: [
+    {
+      name: "position",
+      type: "position",
+    },
+  ],
+  metadataMappings: {
+    street: { type: "keyword" },
+  },
```
afondard, line 12: As the position is a constant it's better in the metadatas (cf [iot-platform](https://github.com/kuzzleio/iot-platform/blob/3-dev/apps/api/lib/modules/public_lighting/assets/StreetLamp.ts))
```suggestion
  measures: [
    {
      name: "brightness",
      type: "brightness",
    },
    {
      name: "powerConsumption",
      type: "powerConsumption",
    },
  ],
  metadataMappings: {
    position: { type: "geo_point" },
    street: { type: "keyword" },
  },
```

afondard (changes requested):

afondard (approved):

PR #370, opened 2024-09-03
feat(metadata): add optional editor hints for metadata

cyrng:
## What does this PR do ?

This PR add new properties on metadetails to allows features to display in the frontend in order to improve the UX. The features are read only mecanism, dropdown of values, calendar picker with time or not and clock picker.
There is an update in the documentation in the concept section to explain how to use it in the frontend.

https://kuzzle.atlassian.net/browse/KZLPRD-473
https://kuzzle.atlassian.net/browse/KZLPRD-474
https://kuzzle.atlassian.net/browse/KZLPRD-475


fmauNeko (dismissed):

etrousset (changes requested):
As seen with Cyril, we should separate different metadata types with spécific edition option is separate objects to avoid ending with a single metaobject with to  many option matrix

lib/modules/model/types/ModelContent.ts
```diff
@@ -33,18 +33,30 @@ export interface LocaleDetails {
...
+  readOnly?: boolean;
+  type?: string;
+  values?: string[] | number[] | boolean[];
+  customValueAllowed?: boolean;
+}
+
+interface DatetimeMetadataDetails {
+  date?: boolean;
+  time?: boolean;
+  timezone?: boolean;
```
fmauNeko, line 46: What happens when timezone is false ? We don't ask for a timezone ?
If that's the case, what timezone will be inferred ? Will we process date and times as UTC or as localtime ? Client or server localtime ?

etrousset: date shouldn't optional in a Datetime picker

cyrng: Yes I understand now that timezone must be always informed to avoid problems through your questionings. I will remove the property and set date to mandatory

fmauNeko (approved):

lib/modules/model/types/ModelContent.ts
```diff
@@ -33,16 +33,27 @@ export interface LocaleDetails {
   description: string;
 }
 
-interface DefinitionMetadataDetails {
+interface BaseDefinition {
   readOnly?: boolean;
-  type?: string;
-  values?: string[] | number[] | boolean[];
+  type: string;
```
fmauNeko, line 38: You can narrow it to `MetadataDetailsEnum`, so it enforces using an enum option when extending `BaseDefinition`.

lib/modules/model/types/ModelContent.ts
```diff
@@ -33,16 +33,27 @@ export interface LocaleDetails {
...
+}
+
+export enum MetadataDetailsEnum {
+  OPTION_SELECTOR = "optionSelector",
+  DATETIME = "datetime",
+}
+
+export interface OptionsSelectorDefinition extends BaseDefinition {
+  type: typeof MetadataDetailsEnum.OPTION_SELECTOR;
```
fmauNeko, line 47: `typeof` is not necessary here, `typeof MetadataDetailsEnum.OPTION_SELECTOR === MetadataDetailsEnum.OPTION_SELECTOR`

lib/modules/model/types/ModelContent.ts
```diff
@@ -33,16 +33,27 @@ export interface LocaleDetails {
...
+  values: string[] | number[] | boolean[];
   customValueAllowed?: boolean;
 }
 
-interface DatetimeMetadataDetails {
+export interface DatetimeDefinition extends BaseDefinition {
+  type: typeof MetadataDetailsEnum.DATETIME;
```
fmauNeko, line 53: Same here about `typeof`

fmauNeko (changes requested):

lib/modules/model/types/ModelContent.ts
etrousset: typeof est pas necessaire, c'est ça?

etrousset (changes requested):
I don't see the documentation about `OptionsSelectorDefinition ` and `DatetimeDefinition` fields to explain de meaning of each option, and how they intented to contrain the frontend editor for the concerned metadata values.

etrousset (approved):

afondard (approved):

fmauNeko (approved):

PR #371, opened 2024-09-04
fix: misleading in documentation data ingestion

EnzoViry1:
## What does this PR do ?

This PR fix documentation about events occuring during data ingestion


Kuruyia (approved):

etrousset (changes requested):
Actualy, the original documentation was right, but misleading in its explanation

PR #374, opened 2024-09-24
fix(device-metadata): fetch device's metadata from assigned tenant when available

cyrng:
## What does this PR do ?

When a payload is sent to register a new measure for a device, the metadata of the device are replaced by those from the index platform (```PayloadService.ts```). So the fix is to retrieve from the associated tenant if it exists. 

KZLPRD-556

  - Step 1 : check the metadata value registered in the device
  - Step 2 : send a payload, for example :
  ```
  {
    "deviceEUI" : "ABE1",
    "lat" : 48.8566,
    "lon" : 2.2523,
    "battery" : 40,
    "externalTemperature" : 1,
    "internalTemperature" : 1
    // "timestamp" : 1711753200000
}
``` 
to this endpoint ```http://localhost:7512/_/device-manager/payload/abeeway``` 

  - Step 3 : check again if the value didn't change

### DIFF
--- a/lib/modules/decoder/PayloadService.ts
+++ b/lib/modules/decoder/PayloadService.ts
@@ -174,6 +174,20 @@ export class PayloadService extends BaseService {
         ),
       );
 
+    // Due to the existence of a "devices" collection in the tenant index and a platform index,
+    // it is need to fetch the device content associated to the tenant if it exists.
+    const updatedDevices = await Promise.all(
+      devices.map((device) =>
+        device._source.engineId && device._source.engineId.trim() !== ""
+          ? this.sdk.document.get<DeviceContent>(
+              device._source.engineId,
+              InternalCollection.DEVICES,
+              device._id,
+            )
+          : device,
+      ),
+    );
+
     // If we have unknown devices, let's check if we should register them
     if (errors.length > 0) {
       const { _source } = await this.sdk.document.get(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/modules/decoder/PayloadService.ts [hunk 1]
```diff
@@ -174,6 +174,20 @@ export class PayloadService extends BaseService {
...
+        device._source.engineId && device._source.engineId.trim() !== ""
+          ? this.sdk.document.get<DeviceContent>(
+              device._source.engineId,
+              InternalCollection.DEVICES,
+              device._id,
+            )
+          : device,
+      ),
+    );
+
```
cyrng, line 190: Doing ```get request``` in a ```map``` does not seems good but if there is a better solution, I'll take it.

lib/modules/decoder/PayloadService.ts
```diff
@@ -174,6 +174,20 @@ export class PayloadService extends BaseService {
         ),
       );
 
+    // Due to the existence of a "devices" collection in the tenant index and a platform index,
+    // it is need to fetch the device content associated to the tenant if it exists.
+    const updatedDevices = await Promise.all(
+      devices.map((device) =>
+        device._source.engineId && device._source.engineId.trim() !== ""
+          ? this.sdk.document.get<DeviceContent>(
```
cyrng, line 182: Doing ```get request``` in a ```map``` does not seems good but if there is a better solution, I'll take it.

etrousset (approved):

PR #377, opened 2024-10-08
fix: trigger events on metadataReplace call

Juiced66:
Adds a triggerEvent to allow piping on `generic:document:afterWrite` as it was possible before KZLPRD-323


fmauNeko (approved):

Kuruyia (approved):

Juiced66:
do not merge please, 2-dev was not aligned when I branched

Issue #353, opened 2024-07-05
[CSV Exporter] File Name Customization


Juiced66:
# Description:
Allow users to specify the filename of the exported CSV. Update the exportMeasures function to accept a fileName parameter and configure the response headers accordingly.

# Tasks:

- Update exportMeasures functions to accept fileName parameter.
- Configure response headers to use the specified fileName.

## Example Code:

```typescript
const fileName = request.getString("fileName", `device-${new Date().toISOString()}.csv`);
request.response.configure({
  headers: {
    "Content-Disposition": `attachment; filename="${fileName}"`,
    "Content-Type": "text/csv",
  },
});
```
## Pros:

- Allows customization of the exported file name.
- Improves user experience by providing meaningful filenames.

## Cons:

- Adds an additional parameter to manage.
- Potential for filename conflicts if not handled properly when requesting.

PR #379, opened 2024-10-27
allow pass customized filename for exporting csv

jackneer:
@Juiced66 
This PR enables a customized filename for exporting devices to csv file #353.
please review, thanks!

### DIFF
--- a/lib/modules/device/DevicesController.ts
+++ b/lib/modules/device/DevicesController.ts
@@ -514,6 +514,7 @@ export class DevicesController {
         const exportId = request.getString("exportId");
 
         const { id } = await this.measureExporter.getExport(engineId, exportId);
+        const filename = request.getString("filename") || `device-${id}.csv`;
         const stream = await this.measureExporter.sendExport(
           engineId,
           exportId,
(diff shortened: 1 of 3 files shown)


Juiced66:
Hi ! Thank you for contributing ! We will review it monday and get back to you !

Juiced66:
Can you please target 2-dev branch, I think we will not release it directly but we can try it in next beta !

Juiced66 (approved):
LGTM

jackneer:
sure, will change target to dev branch

jackneer:
@Juiced66 the target of PR has been changed to 2-dev. please check, thanks!

lib/modules/device/DevicesController.ts
```diff
@@ -514,14 +514,15 @@ export class DevicesController {
         const exportId = request.getString("exportId");
 
         const { id } = await this.measureExporter.getExport(engineId, exportId);
+        const filename = request.getString("filename") || `device-${id}.csv`;
```
rolljee, line 517: I feel like we should enforce the .csv on our end and not let the user decide what extension is being used here to avoid issues.

rolljee (changes requested):

lib/modules/device/DevicesController.ts
jackneer: @rolljee 
I think this depends on how UI communicates with users. we can go even further by detecting filename to check if it contains extension also checking for extension. I'd update PR if that's the requirement.

PR #380, opened 2024-11-12
fix: getAsset should search in commons assets too

thomas-mauran:
## What does this PR do ?

The bug appears when someone tries to create an asset with a model which uses an engineGroup sets to commons, resulting in an error in the frontend saying "insufficient permissions".

The problem comes from the assetCreate in KDM which calls the models:asset:get to get the model of the asset we want to create, the thing is that we don't check if in the engineId: commons, so when this is the case it result in an error being thrown saying the asset we are looking for doesn't exist

  - Step 1 : cloen and run the iot platform
  - Step 2 : run the fixtures
  - Step 3 : modify the model container to change engineId to "commons"
  - Step 4: try to create an asset of type Container

### DIFF
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -23,12 +23,7 @@ module.exports = {
     },
     {
       name: "2-dev",
-      prerelease: "dev",
-      range: "2.x",
-    },
-    {
-      name: "next",
-      prerelease: true,
-    },
+      prerelease: "dev"
+    }
   ],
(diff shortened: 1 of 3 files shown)


Juiced66 (approved):
well done gj

Juiced66 (approved):

sebtiz13 (approved):

PR #382, opened 2024-11-22
feat(deviceManagerEngine): free devices after engine deletion

QuentinRousselet:
## What does this PR do ?
Unassign all the devices from their engineId and assetId in the admin index when the deviceManagerEngine is deleted **after** the deletion of its collections.

  - Step 1 : Create a device-manager-engine
  - Step 2 : Add attach several devices to it and assign them to some assets
  - Step 3 : Delete the engine
  - Step 4 : Check in the admin index in the device collection that all former devices don't have any engineId or assetId anymore.

KZLPRD-511

### DIFF
--- a/lib/modules/device/DeviceService.ts
+++ b/lib/modules/device/DeviceService.ts
@@ -81,44 +81,6 @@ export class DeviceService extends DigitalTwinService {
       },
     );
 
-    /**
-     * On deletion of a tenant, remove association for all of
-     * its formerly linked devices in the admin index.
-     */
-    this.app.hook.register(
-      "multi-tenancy/tenant:afterDelete",
-      async (request) => {
-        const { group, name } = request.input.args;
-        const tenantIndex = `tenant-${group}-${name}`;
-        const devices = [];
-
-        let result = await this.sdk.document.search(
-          this.config.adminIndex,
-          "devices",
-          {
-            _source: false,
-            query: { bool: { must: { term: { engineId: tenantIndex } } } },
-          },
-          {
-            scroll: "2s",
-            size: 100,
-          },
-        );
-        while (result !== null) {
-          devices.push(...result.hits);
-          result = await result.next();
-        }
-        void this.sdk.document.mUpdate(
-          this.config.adminIndex,
-          "devices",
-          devices.map((device) => {
-            return { _id: device._id, body: { assetId: null, engineId: null } };
-          }),
-        );
-      },
-    );
-  }
-
   /**
    * Create a new device.
    *
--- a/lib/modules/device/DeviceService.ts
+++ b/lib/modules/device/DeviceService.ts
@@ -81,7 +81,7 @@ export class DeviceService extends DigitalTwinService {
       },
     );
   }
-  
+
   /**
    * Create a new device.
    *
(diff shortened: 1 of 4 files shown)


lib/modules/device/DeviceService.ts
```diff
@@ -80,6 +80,43 @@ export class DeviceService extends DigitalTwinService {
         await this.attachEngine(engineId, deviceId, request);
       },
     );
+
+    /**
+     * On deletion of a tenant, remove association for all of
+     * its formerly linked devices in the admin index.
+     */
+    this.app.hook.register(
```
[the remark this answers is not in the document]
QuentinRousselet: Indeed, this was the first implementation but supposed to be removed for the reason you gave
My bad, I messed up my last commit...

lib/modules/plugin/DeviceManagerEngine.ts
```diff
@@ -654,4 +653,40 @@ export class DeviceManagerEngine extends AbstractEngine<DeviceManagerPlugin> {
...
     return result.hits.map((elt) => elt._source);
   }
+
+  /**
+   * Detach all devices of an index in the admin index
+   *
+   * @param engineId The target engine Id
+   * @returns {any}
+   */
+  private async detachDevicesFromAdminIndex(engineId: string) {
```
etrousset, line 663: I feel like this method should be named `detachDevicesFromTenantIndex`

etrousset (changes requested):

lib/modules/plugin/DeviceManagerEngine.ts
QuentinRousselet: I think tenantIndex could lead to confusion as this functions is related only to the 'platform' index which is referred as adminIndex in KDM.

Juiced66:
It could be so cool to implement a dependancies property at model level that registers cleaning hooks itself. That way user can register is own dependancies and clean all business deps + system deps

rolljee:
> It could be so cool to implement a dependancies property at model level that registers cleaning hooks itself. That way user can register is own dependancies and clean all business deps + system deps Could be done for all entities of KDM

Could you elaborate on this ? i think i didn't get it properly

Juiced66:
> > It could be so cool to implement a dependancies property at model level that registers cleaning hooks itself. That way user can register is own dependancies and clean all business deps + system deps Could be done for all entities of KDM
> 
> Could you elaborate on this ? i think i didn't get it properly

It could be a kind of service wich we can pass a trigger (create, delete ...) and a callback wich is invoke when the action is trigger. That way we structurate system cleaning and we will be able to easily open it to user with a registerDependancy method 
The idea is if an asset for example should add a workflow on create, we can register there dependancy and register also a cleaning callback when asset is removed for example.

that's fully possible to do it manually for now but it could structure the process and be easier for developpers IMHO

rolljee (changes requested):
You need to add a test for this :) otherwise look good to me

rolljee (commented):
What is the state of this PR ?

QuentinRousselet:
> What is the state of this PR ?

Waiting for other reviews now that I added tests as requested

sebtiz13 (approved):

etrousset (approved):

etrousset:
@rolljee now that the test is implémented, can you approve this PR?

etrousset (changes requested):
The PR should be merged in 2-dev

rolljee (approved):

etrousset (approved):

PR #383, opened 2024-11-22
feat(softTenants): add softTenant ids to assets measures documents

thomas-mauran:
## What does this PR do ?

Add `softTenant` fields to digitalTwins documents

## Other changes

- Import types derivative of `ModelDefinition` like `AssetModel`, `DeviceModel` and `MeasureModel` from iot-platform
- Refactor some tests

## Boyscout

- Improve types export
- Improve tests models definitions to simplify list to clear models for tests
- Add basic VSCode settings


afondard:
Seems like there is a mapping issue because the `dynamic` property of the `assets` collection is set to `strict`, not allowing a new property like `softTenants`.

thomas-mauran:
> Seems like there is a mapping issue because the `dynamic` property of the `assets` collection is set to `strict`, not allowing a new property like `softTenants`.

Yes this is still in draft and not working yet, I didn't had the time to finish

etrousset (changes requested):
This feature should be done on `2-dev` branch

lib/modules/asset/types/AssetApi.ts
```diff
@@ -61,7 +67,8 @@ export type ApiAssetMetadataReplaceResult = KDocument<AssetContent>;
 export interface ApiAssetUpsertRequest extends AssetsControllerRequest {
   action: "upsert";
 
-  _id: string;
+  // TODO it's unused in action, maybe should remove it
+  _id?: string;
```
etrousset, line 71: If this is scritly internal structure and not used, shouldn't we just remove now?

etrousset (approved):

lib/modules/asset/types/AssetApi.ts
sebtiz13: It's seems to be not used. In any case, it's not used in the action, so even if it passed in request it's useless.
I have verified on all GitHub repository and I don't see any usage of this type, so it seems currently not used by ours projects.
IMHO, yes we can remove it now, because that don't create breaking change of behavior, but if this type it's used somewhere it will be necessary to update the code to fix the type error thrown.

etrousset: Let's remove it then

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "javascript.preferences.quoteStyle": "double",
+  "exportall.config.quote": "\"",
+  "exportall.config.relExclusion": [
+    "/lib/modules/shared/exports.ts",
+    "/lib/modules/shared/types/exports.ts"
+  ]
+}
```
afondard, line 8: Is it normal for this file to be there?

afondard (approved):

lib/modules/asset/collections/assetsMappings.ts
```diff
@@ -25,7 +25,10 @@ export const assetsMappings: CollectionMappings = {
         date: { type: "date" },
       },
     },
-
+    softTenant: {
```
afondard, line 28: So there could be only 1 `softTenant`?

.vscode/settings.json
sebtiz13: Yes, It's useful to share basic VSCode settings.
Especially on certain repositories like this one which still have a specific configuration like double quote style instead of simple quote.

Maybe we should rationalize these types of configurations between project but that need a huge PR to reformat all code, to accepted style. So for the moment IMHO, it's the better solution to share these settings for contributors

lib/modules/asset/types/AssetApi.ts
sebtiz13: Ok, it's fixed on type and documentation for asset and devices (same mistake)

lib/modules/asset/collections/assetsMappings.ts
sebtiz13: No, but it's currently implemented like that and change this need a migration of database so it's breaking if we change to pluralize now

lib/modules/device/DevicesController.ts
```diff
@@ -64,9 +64,7 @@ export class DevicesController {
         },
         upsert: {
           handler: this.upsert.bind(this),
-          http: [
-            { path: "device-manager/:engineId/devices/:_id", verb: "post" },
-          ],
+          http: [{ path: "device-manager/:engineId/devices", verb: "put" }],
```
fmauNeko, line 67: This is a breaking API change, this should be marked as such in the commits and lead to a semver-major release.

fmauNeko (approved):

PR #387, opened 2024-12-26
fix: lastMeasuredAt

thomas-mauran:
This was creating a bug where if a payload with an older measure was received we would set lastMeasuredAt to 0


afondard (approved):

rolljee (changes requested):
You should add a test for this case :)

rolljee:
You need to remake this PR.

Start from 2-dev instead of 3-dev

rolljee (changes requested):
See comment

thomas-mauran:
Good brannch is #399

PR #388, opened 2025-01-08
Kzlprd 724 etq utilisateur jvp cr er des rapport type par soft tenants et par groupes

cyrng:
## What does this PR do ?

It adds optional `type` in `AssetGroup` for asset group dashboard template feature

  - Step 1 : include KDM in the `iot-platform`
  - Step 2 : create an asset group for example with this payload
```
  {
  "controller": "device-manager/assetsGroup",
  "action": "create",
  "engineId": "tenant-sdef-sdef",
  "body": {
    "children": [],
    "name": "test groupe 3",
    "parent": null,
    "type": "Batiment"
  }
}
```
- Step 3 : update an asset
- Step 4 : test without `type`

### DIFF
--- a/lib/modules/asset/AssetsGroupsController.ts
+++ b/lib/modules/asset/AssetsGroupsController.ts
@@ -257,7 +257,7 @@ export class AssetsGroupsController {
         lastUpdate: Date.now(),
         name: body.name,
         parent: body.parent ?? null,
-        type: body.type ?? "",
+        type: body.type ?? null,
       },
       _id,
     );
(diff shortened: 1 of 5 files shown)


EnzoViry1 (approved):

EnzoViry1:
LGTM

lib/modules/asset/AssetsGroupsController.ts
```diff
@@ -257,6 +257,7 @@ export class AssetsGroupsController {
         lastUpdate: Date.now(),
         name: body.name,
         parent: body.parent ?? null,
+        type: body.type ?? "",
```
afondard, line 260:
```suggestion
        type: body.type ?? null,
```

afondard (approved):

lib/modules/asset/AssetsGroupsController.ts
afondard: I think initializing it to `null` is better

PR #391, opened 2025-01-10
feat: clean device on detach

thomas-mauran:
clean the device of his lastMeasure, measures {} and remove the document in the measure collection when we detach a device


rolljee (changes requested):
You need to start from 2-dev

thomas-mauran:
fixed in #400

PR #398, opened 2025-01-29
fix(digitalTwin): correct race condition

sebtiz13:
## What does this PR do ?

Add forgotten locks to fix race condition errors on digital twins

### DIFF
--- a/lib/modules/device/DeviceService.ts
+++ b/lib/modules/device/DeviceService.ts
@@ -430,6 +430,59 @@ export class DeviceService extends DigitalTwinService {
   }
 
   /**
+   * Internal logic to attach the device to an engine
+   *
+   * @param engineId Engine id to attach to
+   * @param deviceId Device id to attach
+   * @param options.refresh Wait for ES indexation
+   * @param options.strict If true, throw if an operation isn't possible
+   */
+  private async _attachEngine(
+    engineId: string,
+    deviceId: string,
+    request: KuzzleRequest,
+  ): Promise<KDocument<DeviceContent>> {
+    const device = await this.getInternalDevices(deviceId);
+
+    if (device._source.engineId) {
+      throw new BadRequestError(
+        `Device "${device._id}" is already attached to an engine.`,
+      );
+    }
+
+    await this.checkEngineExists(engineId);
+
+    device._source.engineId = engineId;
+
+    const [updatedDevice] = await Promise.all([
+      this.updateDocument<DeviceContent>(request, device, {
+        collection: InternalCollection.DEVICES,
+        engineId: this.config.adminIndex,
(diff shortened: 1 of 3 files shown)


etrousset (approved):

lib/modules/device/DeviceService.ts
```diff
@@ -192,35 +192,37 @@ export class DeviceService extends DigitalTwinService {
     metadata: Metadata,
     request: KuzzleRequest,
   ): Promise<KDocument<DeviceContent>> {
-    const device = await this.get(engineId, deviceId, request);
+    return lock(`device:${engineId}:${deviceId}`, async () => {
```
Olive3034, line 195: Because you can have no engineId attahed to this device, and deviceId is unique in the whole platform, I guess it should be only :
`return lock(`device:${deviceId}`, async () => {`

And need to modify accordingly the create function

lib/modules/device/DeviceService.ts
```diff
@@ -299,7 +301,7 @@ export class DeviceService extends DigitalTwinService {
     request: KuzzleRequest,
   ): Promise<KDocument<DeviceContent>> {
     return lock<KDocument<DeviceContent>>(
-      `device:update:${deviceId}`,
+      `device:${engineId}:${deviceId}`,
```
Olive3034, line 304: Same comment here

lib/modules/device/DeviceService.ts
```diff
@@ -336,7 +338,7 @@ export class DeviceService extends DigitalTwinService {
     deviceId: string,
     request: KuzzleRequest,
   ) {
-    return lock<void>(`device:delete:${deviceId}`, async () => {
+    return lock<void>(`device:${engineId}:${deviceId}`, async () => {
```
Olive3034, line 341: Same comment here

lib/modules/device/DeviceService.ts
```diff
@@ -542,7 +544,7 @@ export class DeviceService extends DigitalTwinService {
     asset: KDocument<AssetContent>;
     device: KDocument<DeviceContent>;
   }> {
-    return lock(`device:linkAsset:${deviceId}`, async () => {
+    return lock(`device:${engineId}:${deviceId}`, async () => {
```
Olive3034, line 547: Same comment here

Olive3034 (changes requested):

Olive3034 (changes requested):
due to successive lock (upsert for example), now it takes up to 5200ms to perform an upsert.
So I guess, in those specific cases, we should unlock the mutex before calling a new successive action that will acquire the mutex ?

Olive3034 (approved):

PR #403, opened 2025-02-14
feat(assetsGroups): add service to handle group logic

QuentinRousselet:
## What does this PR do ?
This Pr adds an `assetsGroupsService` to handle group logic.
This allows more consistency with device and assets services and permit the use of base service methods allowing for pipes to b


afondard (approved):
LGTM

PR #410, opened 2025-03-03
feat!: remove kdm measures export

Juiced66:
KDM now don't export / auto register measures at plugin start. Responsability is now for users. Avoid conflicts on fields where we already did register and when user define a different mapping


lib/modules/asset/types/AssetHistoryContent.ts
```diff
@@ -8,14 +8,6 @@ export type AssetHistoryMetadata = {
   names: string[];
 };
 
-export type AssetHistoryEventMeasure = {
```
Juiced66, line 11: This change should have been made onto #409 , my bad

fmauNeko (approved):

etrousset (approved):

PR #414, opened 2025-03-10
fix: use original event with new payload

Juiced66:
Refactor events, add docs


lib/modules/measure/MeasureService.ts
```diff
@@ -113,14 +113,14 @@ export class MeasureService extends BaseService {
      *
      * Useful to enrich measures before they are saved.
      */
-    await this.app.trigger<EventMeasureProcessSourceBefore>(
-      "device-manager:measures:process:sourceBefore",
+    await this.app.trigger<EventMeasureProcessBefore>(
```
rolljee, line 116: If i am not mistaken this is a breaking change ? 🤔

etrousset: This is for next major version so this is the time

etrousset (approved):

PR #416, opened 2025-03-25
feat: implement endpoint to retrieve payload with id

cyrng:
## What does this PR do ?

This PR add an endpoint to retrieve payload with the uuid to display its information in the iot platform (for platform admin only)
Related **Iot-Platform PR** : https://github.com/kuzzleio/iot-platform/pull/577
image

image

  - Step 1 : Try to get a payload with  
```
{
  "controller": "device-manager/rawPayloads",
  "action": "get",
  "_id": "ef78cb29-e94f-4f95-b627-cc1aa82ecac5"
}
```
  - Step 2 :
  - Step 3 :


etrousset (changes requested):
* You can put this feature in payloads contrôller as it is the contrôller for the decoders endpoint.
Thus, you should add a device-manager/raw-payloads controller

* Also, this should include user documentation

inkedsquid (approved):

lib/modules/rawPayloads/RawPayloadsController.ts
```diff
@@ -0,0 +1,29 @@
+import { ControllerDefinition, KuzzleRequest } from "kuzzle";
```
Olive3034, line 1: as we can only get only one raw payload (passing a unique id), why using rawPayload**s** instead of rawPayload ?

PR #417, opened 2025-04-14
fix(assetModels): error on create asset model with commons engine

QuentinRousselet:
## What does this PR do ?
This PR fixes the issue on the creation of asset models with the engine group 'commons'.
The model was created but the client receives a 400 error.

The issue was that after creating/updating an asset model we then fetch the assets from this model in the different engine to update them.

In case of commons engine group we must do this for every engine. If no engine is present we don't need to throw an error.

Create an asset model with common engine, check that there are no errors
Create an asset of this model in 2 different engines
Recreate the model with an additional metadata field
Check that your assets have been updated accordingly


afondard (approved):
LGTM

inkedsquid (approved):

rolljee (changes requested):
This should be tested

etrousset (approved):

PR #424, opened 2025-07-07
feat(groups): rework asset groups into broader groups

QuentinRousselet:
## What this PR does ? 
This PR refont the assetsGroups into Groups. 
Groups don't have parents and children anymore but a path containing it's ancestors.
```ts
{
  name: string;
  model?: string | null;
  path: string;
  lastUpdate: number;
  metadata?: TMetadata;
}
groupId = group-electric-boltzmann-70376
{
  "lastUpdate": 1752227458910,
  "metadata": {
    "geolocation": {
      "lat": 12,
      "lon": 12
    },
    "width": 25,
    "zone": "mall"
  },
  "model": "Parking",
  "name": "Odysseum",
  "path": "group-milky-yeats-33653.group-boundless-carpenter-45429.group-electric-boltzmann-70376",
}
```

## Other changes
This PR reallign the 3.0.0 on the evolutions of the 2.0.0


Juiced66 (approved):
Looks good to me, hard to review but 
- test are passing in ci and locally
- no v3 changes has been overwrote

For the feature itself it's ok, for the UX we should have a migration guide (or script ideally) at least for people that will need to migrate their groups to the new format in the new collection.

inkedsquid (approved):

PR #428, opened 2025-07-18
feat(group):Allow devices group and add groupModel affinity

QuentinRousselet:
## What this PR does ? 
Allows group to contain devices on top of assets

Adds an affinity object in the group model

```ts
{
  /**
   * Type of elements accepted in the group
   */
  type: Array<"assets" | "devices">;
  /**
   * Models accepted in the group
   */
  models: {
    assets?: string[];
    devices?: string[];
  };
  /**
   * Weather the affinities are a strict restriction or not
   */
  strict: boolean;
}
```
This affinity define if a group can contain assets and/or devices
It also defines an array of preferred models for each of those types. If the strict field is set to true, the group can only contain assets/devices of those models.  

### Other changes
Adds a listItems endpoint to the group controller to fetch the devices and assets

### DIFF
--- a/lib/modules/asset/types/AssetContent.ts
+++ b/lib/modules/asset/types/AssetContent.ts
@@ -32,7 +32,7 @@ export interface AssetContent<
     measureNames: Array<{ asset: string; device: string; type: string }>;
   }>;
   /**
-   * Id's of asset groups
+   * Path's of asset groups
    */
   groups: Array<{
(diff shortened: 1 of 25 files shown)


Juiced66 (approved):
PR name should be change to be meaningful and represents the work that's done. 

LGTM

afondard (approved):
LGTM

PR #438, opened 2025-09-16
fix(assetsController): await a promise inside of a try-catch

QuentinRousselet:
## What does this PR do ?
A promise was not awaited inside of a try catch


etrousset (approved):

cyrng (approved):
LGTM

PR #439, opened 2025-09-19
fix(assetsController): await promise in try catch

QuentinRousselet:
## What does this PR do ?
A promise was not awaited in a try catch in `AssetsController` `updateModelLocales`


cyrng (approved):
LGTM

PR #441, opened 2025-10-01
fix(measures): add device origin groups field to measure documents

QuentinRousselet:
## What does this PR do ?
This PR adds missing origin device groups in measure documents during ingestion.

### Other changes
Linked to issue [KZLPRD-1039](https://kuzzle.atlassian.net/browse/KZLPRD-1040) prevent the ingestion with no measures

[KZLPRD-1039]: https://kuzzle.atlassian.net/browse/KZLPRD-1039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/modules/decoder/PayloadService.ts
+++ b/lib/modules/decoder/PayloadService.ts
@@ -115,7 +115,14 @@ export class PayloadService extends BaseService {
     for (const device of devices) {
       const {
         _id,
-        _source: { reference, model, metadata, linkedMeasures, engineId },
+        _source: {
+          reference,
+          model,
+          metadata,
+          linkedMeasures,
+          engineId,
+          groups,
+        },
       } = device;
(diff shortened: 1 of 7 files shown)


etrousset (changes requested):
* make a function to fill the source object to avoid copying the same code several time and allow future field addition with avoiding forgetting some places.

Repeated code : 
`source: {
                deviceMetadata: metadata,
                groups,
                id: _id,
                metadata: deviceMetadataChanges,
                model: model,
                reference: reference,
                type: "device",
              },`
* Implement the TODO and remove se code (should be a matter of using the same source object build function

etrousset (approved):

PR #453, opened 2025-12-11
fix(unlinkAsset): throw if no measure to unlink is provided

QuentinRousselet:
## What does this PR do ?
This PR adds a guard in `AssetsController` `unlinkDevices` and `DevicesController` `unlinkAssets` that throws if threre are no measures provided to be unlinked

Try unlinking measures from a device or an assets with an empty array of measures

KZLPRD-1125


etrousset (changes requested):
Shouldn't the error case be described in the API documentation?

etrousset (approved):

PR #468, opened 2026-03-30
fix(models): return all asset models for super admin without engineGroups (KZLPRD-925)

pvermeil-kuzzle:
## What does this PR do ?
Fixes `listAssets` returning empty results when a super admin calls it without `engineGroups`.

### Changes
- Add `listAllAssets()` to `ModelService` — queries all asset models from the models collection without scope filtering
- In `ModelsController.listAssets()`, detect admin profile and call `listAllAssets()` when `engineGroups` is empty
- Non-admin users are unaffected — they still require `engineGroups` for scoped access

- As super admin, call `device-manager/models:listAssets` without `engineGroups` — should return all asset models
- As non-admin, same call should return only commons models (default behavior)

### DIFF
--- a/plugin/lib/modules/model/ModelService.ts
+++ b/plugin/lib/modules/model/ModelService.ts
@@ -678,6 +678,21 @@ export class ModelService extends BaseService {
     );
   }
 
+  /**
+   * List all asset models regardless of engine group scope.
+   * Used for super admin global view.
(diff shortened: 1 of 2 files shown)


pvermeil-kuzzle:
Reopening under correct ticket KZLPRD-1205

PR #469, opened 2026-03-30
fix(models): return all asset models for super admin without engineGroups (KZLPRD-1205)

pvermeil-kuzzle:
## What does this PR do ?
Fixes `listAssets` returning empty results when a super admin calls it without `engineGroups`.

### Changes
- Add `listAllAssets()` to `ModelService` — queries all asset models from the models collection without scope filtering
- In `ModelsController.listAssets()`, detect admin profile and call `listAllAssets()` when `engineGroups` is empty
- Non-admin users are unaffected — they still require `engineGroups` for scoped access

- As super admin, call `device-manager/models:listAssets` without `engineGroups` — should return all asset models
- As non-admin, same call should return only commons models (default behavior)

### DIFF
--- a/plugin/lib/modules/model/ModelService.ts
+++ b/plugin/lib/modules/model/ModelService.ts
@@ -678,6 +678,21 @@ export class ModelService extends BaseService {
     );
   }
 
+  /**
+   * List all asset models regardless of engine group scope.
+   * Used for super admin global view.
+   */
+  async listAllAssets(): Promise<KDocument<AssetModelContent>[]> {
+    const result = await this.sdk.document.search(
+      this.config.platformIndex,
+      InternalCollection.MODELS,
+      { query: { term: { type: "asset" } } },
+      { size: 5000, sort: [{ "asset.model": "asc" }] },
+    );
+
+    return result.hits as unknown as KDocument<AssetModelContent>[];
+  }
+
   async listAsset(
     engineGroups: string[],
     engineId?: string,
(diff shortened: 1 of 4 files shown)


pvermeil-kuzzle:
@etrousset The fix is deployed on PR branch. Here's how to verify:

```bash
# Login as admin
JWT=$(curl -s -X POST http://localhost:7512/_login/local \
  -H 'Content-Type: application/json' \
  -d '{"username":"test-admin","password":"password"}' \
  | python3 -c "import json,sys; print(json.load(sys.stdin)['result']['jwt'])")

# List all asset models as admin — no engineGroups needed
curl -s -X POST http://localhost:7512/_query \
  -H "Authorization: Bearer $JWT" \
  -H 'Content-Type: application/json' \
  -d '{"controller":"device-manager/models","action":"listAssets"}' \
  | python3 -m json.tool
```

**Before:** returned `total: 0` without `engineGroups`.
**After:** returns all asset models across all groups (commons, air_quality, public_lighting, etc.).

Additional behaviors implemented:
- Non-admin without `engineGroups`/`engineId` → `400 Bad Request` with "Missing argument" message
- Non-admin with invalid `engineId` → `404 Not Found`
- Non-admin with invalid `engineGroups` → `404 Not Found`
- Non-admin without access to requested engine → `403 Forbidden` (not 401)
- Commons models always included alongside requested groups

etrousset (approved):

PR #473, opened 2026-04-13
feat(groups): added endpoint for moveGroup action

TheBlizZar:
- Added  endpoint for moveGroup action
- Removed path update from all others updates method


QuentinRousselet (commented):
You should add some documentation for the new endpoint 
(doc/3/controllers/groups)

QuentinRousselet (approved):

doc/3/controllers/groups/moveGroup/index.md
```diff
@@ -0,0 +1,81 @@
...
+### Other protocols
+
+```js
+{
+  "controller": "device-manager/groups",
+  "action": "moveGroup",
+  "engineId": "<engineId>",
+  "_id": "<group id>",
+  "body": {
+    "targetGroupId": "<target group id>"  // use "__root__" to move to root
```
etrousset: We should use `null` instead of a custom magic string

etrousset (changes requested):

etrousset (approved):

PR #484, opened 2026-08-03
feat(MeasuresAdapters): Add measure adapters

TheBlizZar:
## Summary

Introduces **Measure Adapters**: expose a device's raw measure as a meaningful asset measure, without writing a dedicated decoder.

- Adapters are registered in code and read-only over the API.
- One adapter, one source measure, one or more target measures (fan-out).
- Assigned per device on a single measure slot — no separate assignment list, no per-device renaming.
- Every adapted measure keeps a trace of the adapter used and the raw source data.

## Mapping fields

**`MeasureAdapterContent`** (adapter definition)

| Field | Purpose |
|---|---|
| `name` |  adapter name |
| `measureModelSource` | Type of the raw input measure (must be `scope: "device"`) |
| `fieldMapping[].measureModelTarget` | Type of the produced output measure (must be `scope: "asset"`); also used as the produced measure's name |
| `fieldMapping[].source` | Path of the field to read from the source measure (e.g. `readings.ch1`) |
| `fieldMapping[].target` | Path of the field to write into the target measure (e.g. `envQuality.humidity`) |

**`DeviceContent.measureSlots[]`** (on a device)

| Field | Purpose |
|---|---|
| `name` / `type` | Measure slot name/type, unchanged |
| `measureAdapter` | *(new, optional)* Name of the adapter assigned to this slot — absent if no adapter is applied |

**`MeasureContent.origin.measureAdaption`** (on a measure produced by an adapter)

| Field | Purpose |
|---|---|
| `adaptor` | Name of the adapter that produced this measure |
| `rawMeasure` | Full raw values of the source measure, before adaptation (traceability) |

**`MeasureModelContent.measure.scope`** *(new)*

| Value | Purpose |
|---|---|
| `"device"` | Raw/technical measure, usable only as an adapter's source |
| `"asset"` *(default)* | Measure usable directly on an asset, or as an adapter's target |


etrousset (changes requested):
Feedback in ticket

PR #491, opened 2026-09-17
docs: remove the v1 and v2 documentation

etrousset:
## What

Removes the `doc/1` and `doc/2` trees (145 files). Only `doc/3`, the documentation of the current major, remains and is maintained from this branch.

## Why

The v1 and v2 documentation described plugin versions that are no longer published from `3-dev`, and had drifted from the code. Keeping them next to `doc/3` made the doc site and the repository misleading about what the plugin actually exposes.

## Checks

- No CI workflow or configuration references `doc/1` or `doc/2`.
- The `.vscode/settings.json` change that was pending locally is deliberately left out of this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


etrousset:
Note for reviewers: `.github/actions/dead-links/action.yml` still links and scans `/official-plugins/device-manager/1/`, but no workflow in this repository uses that composite action and its `npm run doc:prepare` step points to a script that no longer exists. It is stale rather than broken by this PR; happy to drop it or retarget it to `/3/` in a follow-up if wanted.

